PR #21, opened 2015-06-21
use sigil s for readability

Joe-noh:
Hi, sorry but this is an nit-picking pull request.

Sigil string is more readable than string which has many backslashes, isn't it?

### DIFF
--- a/test/dogma/rules/trailing_whitespace_test.exs
+++ b/test/dogma/rules/trailing_whitespace_test.exs
@@ -36,9 +36,9 @@ defmodule Dogma.Rules.TrailingWhitespaceTest do
 
   with "long lines in triple quote strings" do
     setup context do
-      source = "\"\"\"\n"
-            <> "1 + 1       \n"
-            <> "\"\"\"\n"
+      source = ~s("""\n)
+            <> ~s(1 + 1       \n)
+            <> ~s("""\n)
       script = source |> Script.parse( "foo.ex" )
       %{
         script: TrailingWhitespace.test( script )


lpil:
Good call! There should be a Rule for this. :)

May I ask how you found my new little project? Also, do you have any feedback? I love linters, but I've never written one before!

Joe-noh:
Thank you for merging!

> May I ask how you found my new little project?

I found your project on "New Packages"; bottom left of [hex.pm](https://hex.pm/).
I sometimes look on there.

> do you have any feedback? I love linters, but I've never written one before!

Unfortunately I have never written linter too. but I think it is good to imitate rubocup or to refer to José Valim's code for now. Followings are rules that come to my mind.

- Using pipe for single-argument-single-function-call is not preferred. write `foo(var)`, not `var |> foo`.
- Vertical alignment of `=`, `|>`, `->`, `do:` and some.
- `CamelCase` for module names.
- `snake_case` for variable, function and module attribute names.
- Checking whether there are `(at)moduledoc`, `(at)spec` and so on.

In the future, should have a feature to configure detailed parameters by `config.exs`.

I guess this is the first linter project for elixir.
I'm counting on you. Thank you!

lpil:
Configuration via config.exs would be easier, but I was thinking that maybe it would be good to use a dotfile like rubocop and jshint, as this would allow you to customise your config globally which makes using the liner outside of a mix project nicer.

I'm currently leaning towards config.exs, in part because there's no obvious way of installing hex packages globally.

Do you have any opinions?

Joe-noh:
I think following design is simple, least astonishment.

- `dogma` is installed as project deps (`only: :dev`).
  - don't need to install globally.
- global setting file is `~/.dogma.exs`.
  - It is pretty good if initializing task like `mix dogma.global.init` is provided.
- `config/confg.exs` imports global setting file at first
  - Configuration which comes after the `import`, local project configuration, overwrites global one.

> there's no obvious way of installing hex packages globally.

Just so you know, it is not impossible. Phoenix does with `mix archive.install`. http://www.phoenixframework.org/v0.13.1/docs/installation.

lpil:
I like your suggestions! And I didn't know Phoenix did that- good to know.

Thank you for your feedback and support :)

PR #32, opened 2015-06-30
Make first argument of mix task a path to test

meadsteve:
Most of the time I'm using tools like this I don't want to run the tests on my test code. Or if I do I'll want to run a different set of rules.

How do you feel about making the first argument given to the mix task restrict the run to a directory path?

### DIFF
--- a/lib/dogma.ex
+++ b/lib/dogma.ex
@@ -2,16 +2,16 @@ defmodule Dogma do
   alias Dogma.Script
   alias Dogma.Formatter
 
-  def run do
-    get_scripts
+  def run(dir_path \\ "") do
+    get_scripts(dir_path)
     |> Formatter.start( Formatter.Simple )
     |> test_scripts( Formatter.Simple )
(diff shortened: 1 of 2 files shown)


lpil:
Hi Steve! Thanks for your contribution.

This is certainly something I want to add before 1.0.0, but it seems I've forgotten to make an issue for it.

I would like to be able to pass either a directory or a file path, and I would like some tests for this functionality. I must admit that I am guilty here as I have neglected to test the existing mix task, I just wanted to knock out a quick task for testing with.

However, since we're in alpha, how do you feel about me merging this in now for convenience, and we'll flesh this out more later?

meadsteve:
@lpil totally agree about the test coverage and option between files/directories. Definitely something I'd want to move in to a module. I thought this might be a good start though for the option parsing.

I was quite keen to start using dogma in my project so I'm fully onboard with this being merged (with the knowledge that it will be changed later.)

lpil:
Great, let's do it then. Be aware I might break this later. ;)

PR #35, opened 2015-06-30
Stop trailing whitespace test failing on \r\n lines

meadsteve:
The trailing whitespace rule fails if any lines are terminated with \r\n. Which is a little confusing as it isn't visible.

Might be worth adding a separate rule to catch any \r\n endings however.

### DIFF
--- a/test/dogma/rules/trailing_whitespace_test.exs
+++ b/test/dogma/rules/trailing_whitespace_test.exs
@@ -29,24 +29,17 @@ defmodule Dogma.Rules.TrailingWhitespaceTest do
     ]
   end
 
-  with "windows line endings are okay" do
+  with "lines terminated windows style, \r\n" do
     setup context do
       source = "   'hello'\r\n"
       <> "'how'\r\n"
       <> "  'are'\r\n"
       <> "      'you?'\r\n"
-      <> "'WINDOWS!' \r\n"
       script = source |> Script.parse( "foo.ex" ) |> TrailingWhitespace.test
       %{ script: script }
     end
 
-    should_register_errors [
-      %Error{
-        rule: Dogma.Rules.TrailingWhitespace,
-        message: "Trailing whitespace detected [10]",
-        position: 5
-      }
-    ]
+    should_register_no_errors
   end
 
   with "long lines in triple quote strings" do
--- a/test/dogma/rules/trailing_whitespace_test.exs
+++ b/test/dogma/rules/trailing_whitespace_test.exs
@@ -29,6 +29,26 @@ defmodule Dogma.Rules.TrailingWhitespaceTest do
     ]
   end
 
+  with "windows line endings are okay" do
+    setup context do
+      source = "   'hello'\r\n"
+      <> "'how'\r\n"
+      <> "  'are'\r\n"
+      <> "      'you?'\r\n"
+      <> "'WINDOWS!' \r\n"
+      script = source |> Script.parse( "foo.ex" ) |> TrailingWhitespace.test
+      %{ script: script }
+    end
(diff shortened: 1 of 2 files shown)


lpil:
Are Windows line endings OK though? ;)

test/dogma/rules/trailing_whitespace_test.exs
```diff
@@ -29,6 +29,26 @@ defmodule Dogma.Rules.TrailingWhitespaceTest do
     ]
   end
 
+  with "windows line endings are okay" do
```
lpil: I don't like the naming of this set of tests.

'with windows line endings are okay it should raise errors' nor 'it should raise errors with windows line endings are okay' read correctly.

test/dogma/rules/trailing_whitespace_test.exs
```diff
@@ -29,6 +29,26 @@ defmodule Dogma.Rules.TrailingWhitespaceTest do
...
+    setup context do
+      source = "   'hello'\r\n"
+      <> "'how'\r\n"
+      <> "  'are'\r\n"
+      <> "      'you?'\r\n"
+      <> "'WINDOWS!' \r\n"
+      script = source |> Script.parse( "foo.ex" ) |> TrailingWhitespace.test
```
lpil: Would we need an example that raise an error? Perhaps this makes it less clear what we are testing here. Perhaps we would want to not use the shouldi matcher for this test, so we can explicitly state why we expect that one error, if we continue to test for it.

lpil:
Are there any situations in which we would want to preserve Windows line endings? If not, why don't we strip these line endings in the `Script.parse` phase so we only have to do this once? :)

meadsteve:
@lpil I think there should be a separate rule telling users not to have windows line endings. So for this reason I don't think it makes sense to strip them out first. Primarily I just don't think it should fail the trailing whitespace rule.

test/dogma/rules/trailing_whitespace_test.exs
meadsteve: :+1: I'll update the name

test/dogma/rules/trailing_whitespace_test.exs
meadsteve: I'll have a look at sorting these tests out later today. I think you're right it should be clearer why that one is bad but the others are okay.

lpil:
I think you're right, could you open an issue for that rule please?

I've not used a windows machine in many years. Is it difficult to use unix style line endings? Or will a default rule like this cause endless frustration?

meadsteve:
I spotted this because I use a windows machine at work (.net workplace). It's quite easy to prevent it normally but I quickly hacked a few files I was using on a new machine and accidentally committed windows line endings.

meadsteve:
@lpil any better?

lpil:
Perfect, thanks!

lpil:
@meadsteve Your name on your git commits seems to vary a bit. Is this intentional? I noticed when running the `bin/update_contributors` script.

meadsteve:
hmmm it's probably down to whether I've made the commits on my work laptop. It has a slightly different config. But I thought it was steve.brazier both times. I might need to take a look at that.

lpil:
Want me to amend your commit with that name?

meadsteve:
I'm not too fussed. If you're happy with the commit history looking a bit odd I'd just leave it.

lpil:
Commit history will be buried, that CONTRIBUTORS file will annoy me forever ;)

lpil:
Looks fine in the github UI as you're still the committer :)

Issue #37, opened 2015-07-01
Rule: Detect windows line endings \r\n


meadsteve:
After discussion in #35

lpil:
Closed by #40

PR #40, opened 2015-07-02
Add rule for windows line endings

meadsteve:
Implements #37


lpil:
Looks great! Thanks for your continued interest.

If you're working on a rule or issue could you comment on that issue? That way we can avoid both working on the same thing. :)

PR #42, opened 2015-07-02
[POC] Config as Rules files

meadsteve:
I just thought I'd put together a proof of concept on how the config could be done as a list of rule tuples (I think this is actually similar to how your code worked in an earlier revision).

I've used the Rules.LineLength rule as an example of how this can handle configuration by adding an arity 2 test function with the second argument being a key-value list.


test/dogma/rules/line_length_test.exs
```diff
@@ -28,4 +28,19 @@ defmodule Dogma.Rules.LineLengthTest do
...
+        String.duplicate( "y",  30 ),
+        String.duplicate( "z", 101 ),
+        ] |> Enum.join( "\n" )
+        |> Script.parse( "foo.ex" )
+        |> LineLength.test(max_length: 120)
+        %{ script: script }
+      end
+
+      should_register_no_errors
+    end
```
lpil: This file is indented strangely.

lib/dogma/rules.ex
```diff
@@ -1,25 +1,21 @@
...
+      {Rules.DebuggerStatement},
+      {Rules.FinalNewline},
+      {Rules.LineLength, max_length: 80},
+      {Rules.LiteralInCondition},
+      {Rules.NegatedIfUnless},
+      {Rules.QuotesInString},
+      {Rules.TrailingBlankLines},
+      {Rules.TrailingWhitespace},
+      {Rules.UnlessElse}
+    ]
```
lpil: I don't like this because I keep forgetting to update the damn thing when a new rule is added.

How about keeping something like the previous method under the hood, and then merging it with the additional metadata in order to make the tuples here?

lib/dogma/rules/line_length.ex
```diff
@@ -6,21 +6,23 @@ defmodule Dogma.Rules.LineLength do
...
   end
 
-  defp check_line({i, line}, script) do
-    len = String.length( line )
-    if too_long?( len ) do
-      err = error(len, i)
-      Script.register_error( script, err )
-    else
-      script
+  def get_length_checker(max \\ 80) do
```
lpil: Why has this function been renamed and made public?

lib/dogma/script.ex
```diff
@@ -30,9 +30,16 @@ defmodule Dogma.Script do
...
-  def run_tests(script, rules \\ nil) do
-    (rules || Rules.list)
-    |> Enum.reduce( script, fn(rule, x) -> rule.test x end )
+  def run_tests(script, rule_module \\ nil) do
+    (rule_module || Rules).list()
+    |> Enum.reduce( script, &run_test/2 )
+  end
+
+  defp run_test({rule, custom_config}, script) do
+    rule.test(script, custom_config)
```
lpil: I like how we explicitly pass in the config, makes it easy to test :)

test/dogma/script_test.exs
```diff
@@ -167,8 +176,8 @@ defmodule Dogma.ScriptTest do
     end
 
     should "run the tests", context do
```
lpil: Can we change this test name (and maybe the with block if required) to explain more clearly what exactly is going on here? i.e. What's the relationship with the custom rules and the CustomRules module.

lpil:
This looks great, thanks Steve. :rocket:

lpil:
Also, it looks like we've got some failing test -> https://travis-ci.org/lpil/dogma/builds/69336195

lib/dogma/rules.ex
lpil: Actually now that I think about it the test that checks whether the right number of rules is returned prevents me from forgetting in future.

meadsteve:
Ah I rebased this branch after realising it was out of date and it looks like I missed adding some new tests causing it to fail CI.

lib/dogma/script.ex
meadsteve: Yeah I was quite pleased with how neatly this fell out of the configuration.

lib/dogma/rules/line_length.ex
meadsteve: renamed because it's now a function that returns the check_line function. Think it needs a better name though.
Public by mistake. I'll update this as it doesn't need to be and shouldn't be public.

lib/dogma/rules.ex
meadsteve: Yeah if we consider this file as the config file for "run everything" then the test that checks the rule count guarantees this.

test/dogma/rules/line_length_test.exs
meadsteve: weird. I'll fix it.

lib/dogma/rules.ex
meadsteve: and again I'm generally in favour of keeping things explicit like this rather than scanning directories and merging. But that's your call.

test/dogma/script_test.exs
meadsteve: How about should "run each test in the provided configuration" ?

lpil: Sounds good. :)

lib/dogma/script.ex
lpil: We're clearly genius programmers.

lib/dogma/rules/line_length.ex
lpil: Ooh, it does. I missed that. It would be nice if we could make that clearer somehow.

lpil:
That looks better.

Do we also want to add a field to the tuple that states whether the rule is enabled or not?

lib/dogma/rules/line_length.ex
meadsteve: I've tackled this problem in a different way by using filter then reduced. What do you think? I think it makes the intent clearer.

meadsteve:
my preference would still to remove tuples rather than have an option to disable. Keeps things a bit simpler and clearer.

lpil:
If we do this, how do we ensure that we add new rules to the various lists?

lib/dogma/rules.ex
```diff
@@ -1,25 +1,23 @@
...
+      {Rules.FinalNewline},
+      {Rules.LineLength, max_length: 80},
+      {Rules.LiteralInCondition},
+      {Rules.ModuleName},
+      {Rules.NegatedIfUnless},
+      {Rules.QuotesInString},
+      {Rules.TrailingBlankLines},
+      {Rules.TrailingWhitespace},
+      {Rules.UnlessElse},
+      {Rules.WindowsLineEndings}
```
lpil: Can we drop the `Rules.` ?

meadsteve: I'd quite like to do this but I'm not actually sure how.

meadsteve:
Regarding forgetting to add rules. For the config that is all rules. This is simple and we've already got the test for it. 

For the other configurations (and I suggest we keep the number of maintained ones small) it'll just be something we'd need to remember to do when a new rule gets added. Ask ourselves "does new rule X belong in list Y".

lib/dogma/rules.ex
lpil: Use `Module.concat/2` to prepend `Dogma.Rules` before calling `.test` from it. :)

lpil:
Cool. That seems like the right thing to do to me.

lib/dogma/rules.ex
meadsteve: That could work but then it would mean that no-one would be able to write their own rule in a different space. So I couldn't have ```MeadSteve.Dogma.CustomRule.AllVariablesStartWithSteve```. I'm not sure what the elixir conventions with namespacing things like this are.

lpil: I don't see any reason why we can't have people namespace their rules under `Dogma.Rules`. Or we could make some mechanism for adding rules in another namespace later.

Either way, let's try and write this as nicely as possible for now. I'm very willing to make breaking changes to the API this early on.

meadsteve:
So I've made a small change to the mix task so that the following is possible:
``` bash
mix dogma --rules Dogma.Rules.Sets.All
```
and this will run all the rules defined by list() in the given module.

meadsteve:
So with this added it's basically how I think the config could/should work. So I could define alternative rule sets for example:
```elixir
defmodule Dogma.Rules.Sets.Windows do
  @moduledoc """
  The module which defines all the rules suitable to run on windows
  Rules to be run are returned by `list/0`
  """

  def list do
    [
      {DebuggerStatement},
      {FinalNewline},
      {LineLength, max_length: 80},
      {LiteralInCondition},
      {ModuleDoc},
      {ModuleName},
      {NegatedIfUnless},
      {QuotesInString},
      {TrailingBlankLines},
      {TrailingWhitespace},
      {UnlessElse}
    ]
  end
end
```
interestingly you can also do things like the following:
``` elixir
defmodule Dogma.Rules.Sets.Windows do
  @moduledoc """
  The module which defines all the rules suitable to run on windows
  Rules to be run are returned by `list/0`
  """

  def list do
    Dogma.Rules.Sets.All.list()
    |> Enum.reject(fn
      ({WindowsLineEndings}) -> true
      (_)                    -> false
    end)
  end
end
```
Though I'm not sure how I feel about this.

lpil:
That's quite interesting. Rules are effectively inheritable here.

Right, looks good, besides the eval thing. If you're happy, could you change that and then squash your commits? It'd be great if you add a summary of what you've done in the commit description.

Also, I'm off on holiday tomorrow, so I may not look at this again for a touch over a week after then.

Thanks :dancer:

meadsteve:
oh and  @lpil I forget to say this is now squished in to a few commits.

lpil:
Boom! Let's go! :dancer:

meadsteve:
Exciting times! Though it looks like you've added a test since I tested this PR.

lpil:
Already fixing it :)

PR #43, opened 2015-07-02
New Rule: ModuleDoc

### LINKED ISSUE #28: Rule: ModuleDoc
Enforce presence of `@moduledoc`s

lpil:
Enforces moduledocs to be present.

Makes this project TOTALLY fail it's own checks... Oops. So, is now a good time to allow people to disable rules? ;)

OR maybe we should allow them to be marked as warnings only? I dunno. I'm not sure I want all modules to need moduledocs. Thoughts @meadsteve ?

Closes #28

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ notifications:
 
 script:
   - mix test
-  - mix dogma
+  # - mix dogma
 
 after_success:
   - mix coveralls.travis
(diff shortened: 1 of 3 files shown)


meadsteve:
Agree about the time to start disabling rules. I think this might be the occasion to start having one or two different sets of rules. At least one for everything and one for the rules we'd want to follow on this project (not sure on a name for this set).

My problem with warnings is that I feel they are either going to be something I want to ignore all the time (i.e. that's not a module that needs a doc) or something that should be fixed in which case warning doesn't seem quite right.

lpil:
I'm going to disable linting on CI for now. A bit annoying, but I don't want to wait for this feature before I can start merging in rules.

PR #58, opened 2015-07-15
Add rule for function arity

### LINKED ISSUE #50: Rule: Argument count
Disallow functions with more than `N` arguments. 4?

meadsteve:
Fixes #50


meadsteve:
hmm can't figure out why the coverage has gone down with this PR. Given that it's 0.06% I'm going with: rounding errors.

lpil:
I think you're right about the coverage. It doesn't offer any logical explanation on the web interface either.

Issue #44, opened 2015-07-04
Allow users to specify whether or not each rule runs for .exs files.


lpil:
It was super handy for eca6b060cbbd93d64d08f1ca22248e637b02203f

lpil:
Is this a good idea? How do we want to do this?

PR #61, opened 2015-07-18
Make skipping exs files available to all tests

meadsteve:
So this seemed like the easiest way to implement #44 .
Basically each rule can now have the config: ```skip_exs_files: true``` and it won't run for exs files.

Thoughts?

### DIFF
--- a/lib/dogma/script.ex
+++ b/lib/dogma/script.ex
@@ -8,6 +8,7 @@ defmodule Dogma.Script do
   alias Dogma.Rules
   alias Dogma.Script
   alias Dogma.Error
+  alias Dogma.Script.Skipper
 
   defstruct path:   nil,
             source: nil,
@@ -40,8 +41,16 @@ defmodule Dogma.Script do
   """
   def run_tests(script, rule_module \\ nil) do
     (rule_module || Rules.Sets.All).list()
+    |> Enum.reject(fn(rule) -> skip_test?(rule, script) end)
     |> Enum.map(&namespace_rule/1)
     |> Enum.reduce( script, &run_test/2 )
+  end
+
+  defp skip_test?({_, config}, script) do
+    script |> Skipper.skip?(config)
+  end
+  defp skip_test?(_, _) do
+    false
   end
 
   defp namespace_rule({rule, custom_config}) do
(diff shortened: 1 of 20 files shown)


lib/dogma/script.ex
```diff
@@ -8,6 +8,7 @@ defmodule Dogma.Script do
   alias Dogma.Rules
   alias Dogma.Script
   alias Dogma.Error
+  alias Dogma.Script.Skipper
```
lpil: This name feels a bit odd to me.

lpil:
I'm not sure what I think about this.

Would it be better to have the user supply a regex rather than just pick exs or not? Is it sensible to ignore files like this?

lib/dogma/script.ex
meadsteve: The name is super odd. It worried me a bit that I couldn't find a name. The following is what the module does:

Decides what files should be run based on a modules configuration.

That sounds to me like a discrete unit of work. Therefore suitable for a module. But I couldn't think of a good name. Any suggestions?

meadsteve:
I'm not keen on user supplied regexes as I feel the configuration should say what the module is doing. Rather than how it's done. I just think the following is very easy to understand:
```elixir
{ModuleDoc, skip_exs_files: true}
```
That said it is less flexible. But then I couldn't think of many use cases where I'd actually want to filter files fully based on a regex. Though doing the above wouldn't stop us adding regex filters later.

lpil:
I can understand letting them pass a predicate so that they can control what is run on what- possibly for gradual linting of large legacy codebases. I'm not sure I really understand the logic of being able to toggle `.exs` files for all rules.

meadsteve:
Well I figured that for this project it made sense to exclude the moduledoc rule for .exs files. So it seemed like it would be likely for similar exceptions to bubble up.

I don't think it would be sensible to exclude .exs files for any of the other rules we *currently* have. However excluding certain files from a rule seems like something globally useful rather than just specific to the moduledoc rule.

lpil:
So if it currently only makes sense for one, why are we doing it for all?

meadsteve:
Fair point. But it just feels like the logic to decide if a rule is executed for an exs file doesn't belong with a specific rule. It seems more general.

Happy for this PR to be closed as the discussion seems to have been useful in any case.

lpil:
Both feel a bit odd. Let's wait until we have more rules, then we will have more data with which to decide if this is a common thing or not.

meadsteve:
I think that's a good idea. I'll close this for now.

PR #62, opened 2015-07-19
Fix function_name rule when functions are named with unquote(_)

meadsteve:
Ran in to some code like
```elixir
  def unquote(some_thing)(args) do
    #...
  end
```
This caused an error as the function_name rule was expecting a string. I've modified the rule so it skips any thing like this.

### DIFF
--- a/lib/dogma/rules/function_name.ex
+++ b/lib/dogma/rules/function_name.ex
@@ -22,6 +22,11 @@ defmodule Dogma.Rules.FunctionName do
     {node, errors}
   end
 
+  # If the function is named by unquoting something then we can't check it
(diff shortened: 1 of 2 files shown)


lpil:
Great! Thanks Steve! :)

Sorry for the delay, super busy at the moment.

meadsteve:
no worries. I've just been running off a branch on my fork anyway. Which reminds me. You might like to know I'm actively using this in my project now:

https://github.com/meadsteve/white-bread/blob/master/.travis.yml#L13

lpil:
Awesome! Possibly Dogma's first (non-me) user :D

PR #63, opened 2015-07-19
Fix for function arity rule to handle defprotocol

meadsteve:
A defprotocol will come up as a :def in the ast but won't match
```{_, [line: line_number], [{_, _, args}, _function_body]}``` as the function body is missing.


lpil:
Good catch, thanks!

PR #68, opened 2015-08-14
Add behaviour for configurable rules

meadsteve:


lpil:
Should we just make all rules list a keyword list of options? It could be empty for most.

meadsteve:
@lpil I like that idea. I'll close this PR in favour of doing that.

PR #69, opened 2015-08-14
Add test/2 to rule behaviour

meadsteve:
I've also removed test/dogma/rules_test.exs as it seems the behaviour check should cover this.


meadsteve:
As discussed in #68

lib/dogma/rules/final_newline.ex
```diff
@@ -8,7 +8,9 @@ defmodule Dogma.Rules.FinalNewline do
   alias Dogma.Script
   alias Dogma.Error
 
-  def test(script) do
+  def test(script), do: test(script, [])
+
+  def test(script, _config = []) do
```
lpil: Shouldn't this be `def test(script, _config \\ [])`?

lib/dogma/rules/final_newline.ex
```diff
@@ -8,7 +8,9 @@ defmodule Dogma.Rules.FinalNewline do
   alias Dogma.Script
   alias Dogma.Error
 
-  def test(script) do
+  def test(script), do: test(script, [])
```
lpil: We don't need this if `test/2` is defined with a default argument in the second position.

lib/dogma/rules/final_newline.ex
meadsteve: I considered this but then I didn't want it to match:
```elixir
test(script, banana: "not_valid_config_key")
```
I wonder if it is it possible/sensible to do
```elixir
def test(script, _config = [] \\ [])?
```

Thoughts @lpil ?

lpil: I'm not sure what that is supposed to do?

Why do you not want that to match?

meadsteve: For final_newline (and in fact most of the rules) we don't have any config that can be provided. So I'd say calling it with anything else other than an empty config list is an error.

lpil: Hm. Makes sense. I wonder if there is a nicer way to do this... I'll have a think.

meadsteve: @lpil yeah I wasn't keen on the fact I've added all this boilerplate code for rules. Currently I haven't thought of anything neater than having another function like valid_config?(config).

lpil: We could get rid of `test/1` and always have rules called with two functions, then it's in one place.

There's a bunch of little refactorings I want to do, but I've just started a new job, so I don't have the time at the moment I'm afraid!

meadsteve: @lpil getting rid of test/1 is quite a nice idea. Planning on doing some work this weekend so I'll take a look then.

meadsteve:
@lpil I've updated (and rebased whilst I was at it) this PR. I've kept test/1 and test/2 in the rule behaviour but I've updated all the rules so that most of them only have a single def like:
```elixir
def test(script, _config = [] \\ []) do
...
end
```

so they only match empty configs and default to an empty config when called as test/1. The later makes the unit tests for the rules read much more cleanly than manually inserting [] everywhere.

What do you think?

lpil:
Marvellous. :)

I feel we will build some argument verifying thing in future, but for now this looks great.

PR #72, opened 2015-08-20
Run rules on a script in parallel

lpil:
This actually seems to negatively impact performance...

Using [this](https://github.com/lpil/dogma/blob/master/bench/single_file_bench.exs) benchmark:

| Bench     | Typical result |
| --------- | -------------- |
| Series    | 204.51 µs/op   |
| Parallel  | 445.52 µs/op   |

I guess we shouldn't run things in parallel when we're not actually doing anything that expensive? I think I may have made an assumption about the amount of computation required to make it break even, much like when I was playing with the `Stream` module.

Or am I doing something wrong?

@meadsteve @jonnyarnold

### DIFF
--- a/lib/dogma/script.ex
+++ b/lib/dogma/script.ex
@@ -50,7 +50,8 @@ defmodule Dogma.Script do
   def run_tests(script, rule_module \\ nil) do
     (rule_module || Rules.Sets.All).list()
     |> Enum.map( &namespace_rule/1 )
-    |> Enum.map( &run_test(&1, script) )
+    |> Enum.map( &Task.async(fn -> run_test(&1, script) end) )
+    |> Enum.map( &Task.await(&1) )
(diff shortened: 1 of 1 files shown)


meadsteve:
My first thought is that it may well be slower when the number of files is small. But once this gets larger and the files get larger I suspect the overhead of parallel will be outweighed by the benefit.

Might be interesting to plot a graph of benchmark against number of files.

I'll take a better look at the PR in the meantime though.

lpil:
Wouldn't we need to be parallelising the scripts, not the rules then?

meadsteve:
@lpil that's a very good point. In my defence it is very early in the morning :smile: 

In that case I agree it might be worth skipping this in favour of paralleling the scripts. Seems unlikely we are ever going to have enough rules to justify this.

Issue #48, opened 2015-07-04
Rule: Predicate name


lpil:
```elixir
# bad
def is_even?

# good
def even?

# bad
def has_value?

# good
def value?
```

lpil:
Might be worth checking Rubocop to see how they do this.

PR #75, opened 2015-08-22
Add predicate check - solves #48

revdan:
I hope this PR matches your code & testing style closely enough!  I had a look at the Rubocop implementation of this rule and found it quite unwieldy. Hopefully the Regex I've used is strong enough.

I also added the `is_something` style into the 'good' list, as there's a few native functions that use it. I have no idea why, but as long as the function name's not trying to be both I figured it was fine. 

:bow:

### DIFF
--- a/lib/dogma/rules/module_doc.ex
+++ b/lib/dogma/rules/module_doc.ex
@@ -21,7 +21,7 @@ defmodule Dogma.Rules.ModuleDoc do
   end
 
   defp check_node({:defmodule, m, [_, [do: module_body]]} = node, errors) do
-    if module_body |> has_moduledoc? do
+    if module_body |> moduledoc? do
       {node, errors}
     else
       {node, [error( m[:line] ) | errors]}
@@ -32,32 +32,30 @@ defmodule Dogma.Rules.ModuleDoc do
   end
 
 
-  defp has_moduledoc?(node) do
-    {_, pred} = Macro.prewalk( node, false, &has_moduledoc?(&1, &2) )
+  defp moduledoc?(node) do
+    {_, pred} = Macro.prewalk( node, false, &moduledoc?(&1, &2) )
     pred
   end
 
-
   # moduledocs inside child modules don't count
-  defp has_moduledoc?({:defmodule, _, _}, found_or_not) do
+  defp moduledoc?({:defmodule, _, _}, found_or_not) do
     {[], found_or_not}
   end
 
   # We've found a moduledoc
-  defp has_moduledoc?({:@, _, [{:moduledoc, _, _} | _]}, _) do
+  defp moduledoc?({:@, _, [{:moduledoc, _, _} | _]}, _) do
     {[], true}
   end
 
   # We've already found one, so don't check further
-  defp has_moduledoc?(_, true) do
+  defp moduledoc?(_, true) do
     {[], true}
   end
 
   # Nothing here, keep looking
-  defp has_moduledoc?(node, false) do
+  defp moduledoc?(node, false) do
     {node, false}
   end
-
 
   defp error(position) do
     %Error{
--- a/lib/dogma/rules/module_doc.ex
+++ b/lib/dogma/rules/module_doc.ex
@@ -31,7 +31,6 @@ defmodule Dogma.Rules.ModuleDoc do
     {node, errors}
   end
 
-
   defp moduledoc?(node) do
     {_, pred} = Macro.prewalk( node, false, &moduledoc?(&1, &2) )
     pred
--- a/lib/dogma/rules/predicate_name.ex
+++ b/lib/dogma/rules/predicate_name.ex
@@ -1,0 +1,52 @@
+defmodule Dogma.Rules.PredicateName do
+  @moduledoc """
+  A rule that disallows tautological predicate names.
+  """
+
+  @behaviour Dogma.Rule
+  alias Dogma.Script
+  alias Dogma.Error
+
+  def test(script) do
+    script |> Script.walk( &check_node(&1, &2) )
+  end
+
+  defp check_node({:def, _, [{name, meta, _}|_]} = node, errors) do
+    test_predicate(name, meta, node, errors)
+  end
+  defp check_node({:defp, _, [{name, meta, _}|_]} = node, errors) do
+    test_predicate(name, meta, node, errors)
+  end
+  defp check_node(node, errors) do
+    {node, errors}
+  end
+
+  defp test_predicate(line) do
+    Regex.run(~r{(is|has)_(\w+)\?}, line)
+  end
+
+  defp tautological_predicate?(line) do
+    !!(test_predicate(line))
+  end
+
+  defp test_predicate({:unquote,_,_} , _meta, node, errors) do
+    {node, errors}
+  end
+
+  defp test_predicate(function_name, meta, node, errors) do
+    name = function_name |> to_string
+    if name |> tautological_predicate? do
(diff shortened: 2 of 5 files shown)


revdan:
Regarding that failed check :arrow_double_up: 

It didn't fail locally when I ran `mix test`; is there anything else I should be doing? 

Secondly, it fails because there are functions inside `lib/dogma/rules/module_doc.ex` that don't pass the non-tautological predicate name test :blush: 

I'll rewrite those in the meantime :)

lpil:
Heya revdan! How's it going? Thanks for this :)

When checking certain function/macro calls (such as `def/2` and `defp/2`) I prefer checking the AST to checking the source/lines as then we can be sure that it is definitely a call to those functions, and not some other code that matches our regex. See the [FunctionName](https://github.com/lpil/dogma/blob/master/lib/dogma/rules/function_name.ex#L17-L25) rule for an example of this.

revdan:
Okay, really sorry. If you can tell me why those last two are failing I will accept defeat and fix it up. Grepping other functions

lib/dogma/rules/module_doc.ex
```diff
@@ -32,29 +32,29 @@ defmodule Dogma.Rules.ModuleDoc do
   end
 
 
-  defp has_moduledoc?(node) do
-    {_, pred} = Macro.prewalk( node, false, &has_moduledoc?(&1, &2) )
+  defp has_moduledoc(node) do
```
lpil: I think these should be `moduledoc?`, as they return a boolean.

revdan:
And thanks for the previous feedback, will look into that too :)

lib/dogma/rules/module_doc.ex
revdan: Makes sense! Will change that.

lib/dogma/rules/predicate_name.ex
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  defp tautological_predicate({_, line}) do
+    Regex.match?(~r{defp?\s(is|has)_\w+\?}, line)
+  end
+
+  defp error({line_num, _}) do
+    %Error{
+      rule:     __MODULE__,
+      message:  "Either `is_predicate` or `predicate?` would be sufficient",
```
lpil: Could we have the message tell us the actual violating function name, and then recommend it with the prefix stripped?

i.e. "Favour `moduledoc?` over `has_moduledoc?`"

test/dogma/rules/predicate_name_test.exs
```diff
@@ -0,0 +1,46 @@
+defmodule Dogma.Rules.PredicateNameTest do
+  use DogmaTest.Helper
+
+  alias Dogma.Rules.PredicateName
+  alias Dogma.Script
+  alias Dogma.Error
+
+  with "bad predicates" do
+    setup context do
+      errors = [
```
lpil: Prefer a `"""` multi line string over a list of strings :)

revdan: I actually did it that way first then shortened it :) Will change.

lib/dogma/rules/predicate_name.ex
revdan: :+1:

lpil:
I suspect that the the code-like strings [here](https://github.com/revdan/dogma/blob/drj-48-predicate-name/test/dogma/rules/predicate_name_test.exs#L11-L12) are being picked up by your regex. This is super annoying!

My preferred solution is to walk the AST instead of using lines, as then we can be sure not to trip up over string contents.

There's also ["processed" fields on the %Script struct](https://github.com/lpil/dogma/blob/master/lib/dogma/script.ex#L16-L17) that contain the source file with all string contents stripped out using the [Dogma.Util.ScriptStrings](https://github.com/lpil/dogma/blob/master/lib/dogma/util/script_strings.ex) module, but I try to avoid that where possible.

revdan:
Okay, hope this is better. I'm not that happy with the new module now, as it seems there's quite a bit of duplication with the example you linked to. Any recommendations on how to improve that are very welcome :)

lib/dogma/rules/predicate_name.ex
```diff
@@ -0,0 +1,52 @@
...
+  end
+  defp check_node({:defp, _, [{name, meta, _}|_]} = node, errors) do
+    test_predicate(name, meta, node, errors)
+  end
+  defp check_node(node, errors) do
+    {node, errors}
+  end
+
+  defp test_predicate(line) do
+    Regex.run(~r{(is|has)_(\w+)\?}, line)
```
lpil: Can we put start and end anchors on this regex? :)

`\A` and `\z`

test/dogma/rules/predicate_name_test.exs
lpil: Oh! I mean like this, no joining needed :)

```elixir
        """
        def is_naughty?(arg) do
          true
        end
        defp is_bad?() do
          true
        end
        """
```

lib/dogma/rules/predicate_name.ex
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  defp tautological_predicate?(line) do
+    !!(test_predicate(line))
```
lpil: Having a function that does nothing but cast to boolean doesn't really add anything in a language with duck typing, so let's remove this, and just use `test_predicate/1` in its place.

lpil:
I think the duplication here is fine. We could make a function name checker utilility module/function that take a predicate and an error generating function, but I don't know if that'll make things nicer.

Either way that can come later, or never. This looks great, thank you. Just those little notes I've made above.

revdan:
Thanks for the advice on all this, and sorry for my frequent boo-boos.

lpil:
Nothing to be sorry about, I'm just picky. (The fact I'm writing a linter may have given that away...)

Good to see that you've run that mix task to generate the rules docs. I wasn't sure it was clear enough.

I'm curious, when doing dev here, did you use the `mix test.watch` task to run the tests?

*Anyway*. I've annoyed you more than is fair. This looks great, I'll merge it in. Really good to have someone contributing to my little project :)

revdan:
Thanks so much! I appreciate all the feedback; hopefully my contribution is idiomatic and doesn't make you cringe too bad in 6 months time :)

I haven't used `mix test.watch` yet but running it just now I get:
```
[Porcelain]: goon executable not found
[Porcelain]: falling back to the basic driver
```
and then it takes ages to run the tests. I'll try and fix that up before taking on another issue as it does seem more streamlined than what I was doing...

And I only noticed the rules docs thing because the build failed and I hadn't realised there'd been an update on master since my last commit. One rebase later and all was okay. .

lpil:
No cringe here at all. :)

Yep, that's a known problem. I think I know how to solve it now, just need to get around to doing it. Dogma keeps distracting me.

Not sure why it's slow. What platform are you on? Is it only slow the first time, or every time?

lpil:
Being a bit nosy here- I see your company is based in London. The Elixir meetup is held in Angel every month or so, maybe check it out if you you're into Elixir and free pizza.

revdan:
I had no idea there was a meetup! When's the next one?

I'm on OSX, and to be fair this is my crappy 2011 iMac, so probably nothing to worry about. I'll give it another try on my MBP tomorrow.

lpil:
Here you go! http://www.meetup.com/elixir-london/

revdan:
Brilliant, thank you! I've signed up.

PR #80, opened 2015-08-22
New Rule: Comparison to Booleans

### LINKED ISSUE #45: Rule: Comparison to boolean
Disallow these

```elixir
foo   ==  false
foo   ==  true
foo   === false
foo   === true
foo   !=  false
foo   !=  true
foo   !== false
foo   !== true
true  ==  foo
true  === foo
true  !=  foo
true  !== foo
false ==  foo
false === foo
false !=  foo
false !== foo
```

Because they're totally pointless.

5thWall:
Enforces no comparisons to booleans.

The repo fails in two places with this test. It's possible the user wants to ensure something is literally `true` or `false` instead of just truthy or falsy. Maybe this should be a warning? Thoughts?

Closes #45

### DIFF
--- a/lib/dogma/rules/comparison_to_boolean.ex
+++ b/lib/dogma/rules/comparison_to_boolean.ex
@@ -1,0 +1,33 @@
+defmodule Dogma.Rules.ComparisonToBoolean do
+  @moduledoc """
+  A rule that disallows comparison to booleans.
+  """
+
+  @behaviour Dogma.Rule
+
+  alias Dogma.Script
+  alias Dogma.Error
+
+  def test(script) do
+    script |> Script.walk( &check_node(&1, &2) )
+  end
+
+  for fun <- [:==, :===, :!=, :!==] do
+    defp check_node({unquote(fun), meta, [lhs, rhs]}, errors)
+    when is_boolean(lhs) or is_boolean(rhs) do
+      {node, [error(meta[:line]) | errors]}
+    end
+  end
+
+  defp check_node(node, errors) do
+    {node, errors}
+  end
+
+  defp error(pos) do
+    %Error{
+      rule:    __MODULE__,
+      message: "Comparison to a boolean is pointless",
+      line:    pos
+    }
+  end
+end
--- a/test/dogma/rules/comparison_to_boolean_test.exs
+++ b/test/dogma/rules/comparison_to_boolean_test.exs
@@ -1,0 +1,110 @@
+defmodule Dogma.Rules.ComparisonToBooleanTest do
+  use DogmaTest.Helper
+
+  alias Dogma.Rules.ComparisonToBoolean
+  alias Dogma.Script
+  alias Dogma.Error
+
+  @errors [
+    %Error{
+      rule:    ComparisonToBoolean,
+      message: "Comparison to a boolean is pointless",
+      line:    8
+    },
+    %Error{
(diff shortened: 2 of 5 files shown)


lib/dogma/rules/comparison_to_boolean.ex
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  defp check_node(node, errors) do
+    {node, errors}
+  end
+
+  defp error(pos) do
+    %Error{
+      rule: __MODULE__,
+      message: "Comparison to a boolean is usually pontless",
```
lpil: It's always pointless. If not we shouldn't have a rule for it.

test/dogma/rules/comparison_to_boolean_test.exs
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  with "vars on both sides" do
+    setup context do
+      errors = for fun <- @comp, into: "" do
+        "foo #{fun} bar\n"
+      end |> test
+      %{errors: errors}
+    end
+
```
lpil: There is so much metaprogramming in this set of tests that it's not possible to tell what we're testing, or what the expected behaviour is.

Tests should be simple, declarative, and easy to understand. Can we re-write this file please? Verbosity is fine.

lpil:
Hey! Thanks for this.

I've made some notes above. I'm especially interested in making the tests exemplary as currently I don't think they are very helpful.

`script.ex` needs modification for the build to pass. :)

5thWall:
I've addressed the issues. Do you think there need to be more test cases added? Comparison with other types?

test/dogma/script_test.exs
```diff
@@ -93,7 +93,7 @@ defmodule Dogma.ScriptTest do
       end
 
       should "assign valid? as false", context do
-        assert false == context.script.valid?
+        assert not context.script.valid?
```
lpil: Swap `assert not` for `refute`. :)

test/dogma/rules/comparison_to_boolean_test.exs
```diff
@@ -0,0 +1,115 @@
...
+    }
+  ]
+
+  def test(script) do
+    script
+    |> Script.parse("foo.ex")
+    |> ComparisonToBoolean.test
+  end
+
+  def errors(count) do
```
lpil: Do we still use this function?

lib/dogma/rules/comparison_to_boolean.ex
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  defp check_node(node, errors) do
+    {node, errors}
+  end
+
+  defp error(pos) do
+    %Error{
+      rule: __MODULE__,
+      message: "Comparison to a boolean is pontless",
```
lpil: Typo, pointless. :)

lpil:
Looks good! Thank you. Just a couple more things.

lpil:
Brilliant, thank you!

5thWall:
Looks like there's an unused alias in `lib/dogma/script.ex`. There's an alias for `Dogma.Rules` but the module is always referenced with the full name. Want me to fix it here or wait for you to fix master and rebase again?

lpil:
That's my completely fault, it's broken in master already. This looks good though, so I'll merge it in and fix the build when I next get access to a computer. Thanks!

PR #82, opened 2015-08-24
New Rule: Enforce bare variables starting fn chain

### LINKED ISSUE #53: Rule: Pipe bare value start
The first argument in a chain of pipes should be a bare value, not a function call with an argument.

I've worded this poorly.

5thWall:
Here's my shot at a rule for enforcing bare variables at the start of a function chain. Though, I'm honestly not sure if I've caught all possible cases. Though, I certainly got everything in this repo.

I didn't try to get the case where function arguments aren't surrounded by parens, since these don't even parse to a proper AST. Since it would take another pass to catch them, I think they should be in another rule.

Closes #53

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /deps
 erl_crash.dump
 *.ez
+*.beam
 /doc/
 /docs/all.json
 /bench/snapshots
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -1,10 +1,11 @@
 # Dogma Rules
 
 These are the rules included in Dogma by default. Currently there are
-18 of them.
+19 of them.
 
 ## Contents
 
+* [BarePipeChainStart](https://github.com/lpil/dogma/blob/master/docs/rules.md#barepipechainstart)
 * [ComparisonToBoolean](https://github.com/lpil/dogma/blob/master/docs/rules.md#comparisontoboolean)
 * [DebuggerStatement](https://github.com/lpil/dogma/blob/master/docs/rules.md#debuggerstatement)
 * [FinalNewline](https://github.com/lpil/dogma/blob/master/docs/rules.md#finalnewline)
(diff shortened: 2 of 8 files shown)


meadsteve:
Looks like the master build is failing because of an unused alias:
```
Compiled lib/dogma/util/cyclomatic_complexity.ex
lib/dogma/script.ex:8: warning: unused alias Rules
```
so the CI error isn't actually anything to do with this PR

lpil:
Yep, I broke it on master yesterday.

lpil:
Can we call something with pipe in the name as it's only for pipes? :)

lib/dogma/rules/bare_chain_start_test.ex
```diff
@@ -0,0 +1,69 @@
...
+  # exception for bare maps
+  defp function_line({:%, _, _}),
+  do: nil
+
+  # exception for structs
+  defp function_line({:%{}, _, _}),
+  do: nil
+
+  # exception for module attributes
+  defp function_line({:@, _, _}),
```
lpil: Do we need an exception for tuples of size greater than 2?

lib/dogma/rules/bare_chain_start_test.ex
```diff
@@ -0,0 +1,69 @@
...
+  defp function_line({{:., meta, _}, _, _}),
+  do: {:ok, meta[:line]}
+
+  defp function_line(_),
+  do: nil
+
+  defp error(line) do
+    %Error{
+      rule:    __MODULE__,
+      message: "Method Chains must start with a bare value",
```
lpil: Functions! No methods in Elixir :)

5thWall:
I'm honestly not sure what would be a good name for this. But it has pipe in the name now.

lpil:
That'll do for now. Thanks for this, good work :)

PR #83, opened 2015-08-26
Remove default prefix from configured ruleset.

meadsteve:
As discussed in commit 8f6eee3

### DIFF
--- a/lib/dogma/rules.ex
+++ b/lib/dogma/rules.ex
@@ -7,6 +7,7 @@ defmodule Dogma.Rules do
   alias Dogma.Formatter
   alias Dogma.Script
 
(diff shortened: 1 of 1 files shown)


lib/dogma/rules.ex
```diff
@@ -7,6 +7,7 @@ defmodule Dogma.Rules do
   alias Dogma.Formatter
   alias Dogma.Script
 
+  @default_rule_set Dogma.RuleSet.All
```
lpil: I really like the use of a module attribute here! :)

meadsteve: Thanks! I love them for this kind of thing as you can use the name to say why it exists.

PR #86, opened 2015-08-28
Paralell scripts

### LINKED ISSUE #85: Run files in parallel
After looking at #71 

Might be some benefit in parallelising the run on each file. The first step would be writing some simple benchmarking for this,

meadsteve:
Fixes #85.

Running the benchmark on my machine [Intel Core i7-4500U CPU @ 1.80GHz X 4] gives the following reductions in runtime

## Dogma.RulesRunBench
30 files    -44.53%
20 files    -39.84%
10 files    -33.26%
5 files     -15.12%
1 file      +25.34%

So after 5 files we get a decent saving.


lib/dogma/rules.ex
```diff
@@ -28,6 +29,9 @@ defmodule Dogma.Rules do
     Application.get_env :dogma, :rule_set, @default_rule_set
   end
 
+  defp async_test_script(script, formatter, rule_set) do
```
lpil: This function doesn't actually do anything. Can we remove it and just use `Task.async/1` directly in the `Enum.map/2`?

bench/single_file_bench.exs
```diff
@@ -1,27 +0,0 @@
-defmodule Dogma.SingleSourceBench do
```
lpil: Do we want to remove this? I don't know.

lpil:
Hey, cool, it's faster. That's pretty exciting.

I've got to say, I really like how easy it is to run things in parallel with Elixir. And I feel you could probably quite easily make something faster than async/await if you wanted to.

bench/single_file_bench.exs
meadsteve: TBH I removed it because it seemed broken and it was easier to delete than fix :smile: If it's useful I could spend some time fixing it.

lib/dogma/rules.ex
meadsteve: Doesn't actively do anything but putting it in the enum call looked like:
```elixir
Enum.map(&Task.async(fn -> test_script(&1, formatter, rule_set) end))
```
which just felt a bit heavy. But happy to change to this if you don't like async_test_script/3

meadsteve:
yup. I really liked how we got this performance boost without any real added complexity.

bench/single_file_bench.exs
lpil: Seems reasonable.

lib/dogma/rules.ex
lpil: I think this looks the same, and would rather have it inline if that's cool.

bench/single_file_bench.exs
meadsteve: and actually @lpil  while I'm thinking about this how do you feel about having the benchmarking hooked up to travis ci job? It'll stop it getting broken again. Plus if we wanted to we could write something to compare each build to see how the performance has changed over time.

lib/dogma/rules.ex
meadsteve: @lpil fixed.

bench/single_file_bench.exs
lpil: It's probably a good idea to test the benchmarks, but I don't think that we can run benchmarks on CI as we don't know if the resources available will be the same on each run.

lpil:
Thanks!

PR #87, opened 2015-08-29
Add emacs formatter (for Syntastic/Flycheck integration)

### LINKED ISSUE #8: Syntastic compatible formatter
And you know, Emacs flycheck too, I guess.

5thWall:
Here's my shot at a Syntastic/Flycheck format. The Syntastic checker for Rubocop uses the emacs format. Both Syntastic and Flycheck require a custom checker to be built for each tool, but this format seems to be the standard output used.

I've added tests for option parsing, but I'm not sure how you want to handle integration tests.

Closes #8

### DIFF
--- a/lib/dogma.ex
+++ b/lib/dogma.ex
@@ -10,13 +10,12 @@ defmodule Dogma do
   alias Dogma.Rules
   alias Dogma.ScriptSources
 
-  def run(dir_path \\ "") do
-    dir_path
+  def run({dir, formatter}) do
+    dir
     |> ScriptSources.find
     |> ScriptSources.to_scripts
-    |> Formatter.start
-    |> Rules.test
-    |> Formatter.finish
+    |> Formatter.start(formatter)
+    |> Rules.test(formatter)
+    |> Formatter.finish(formatter)
   end
-
 end
--- a/README.md
+++ b/README.md
@@ -62,6 +62,50 @@ Inspecting 27 files.
 
 How handy!
 
+### Formatters
+
+You can pass a format to the mix task using the `--format` flag.
+
+```
+> mix dogma --format=flycheck
+
+lib/dogma/rules.ex:23:1: W: Blank lines detected at end of file
+test/dogma/formatter_test.exs:9:1: W: Trailing whitespace detected
+```
+
+#### Simple
+
+This is the default formatter. It displays the progress of the checker
+followed by a summary of violations.
+
+```
+> mix dogma --format=simple
+
+Inspecting 27 files.
+
+.....X..........X..........
+
+27 files, 2 errors!
+
+== lib/dogma/rules.ex ==
+23: TrailingBlankLines: Blank lines detected at end of file
+
+== test/dogma/formatter_test.exs ==
+9: TrailingWhitespace: Trailing whitespace detected [33]
+```
+
+#### Flycheck
+
+A machine-readable format suitable for integration with tools like
+[Flycheck](https://github.com/flycheck/flycheck) or
+[Syntastic](https://github.com/scrooloose/syntastic).
+
+```
+> mix dogma --format=flycheck
+
+lib/dogma/rules.ex:23:1: W: Blank lines detected at end of file
+test/dogma/formatter_test.exs:9:1: W: Trailing whitespace detected
+```
 
 ## Contributor Information
 
--- a/lib/dogma/formatter/flycheck.ex
+++ b/lib/dogma/formatter/flycheck.ex
@@ -1,0 +1,48 @@
+defmodule Dogma.Formatter.Flycheck do
+  @moduledoc """
+  A formatter that lists information for files with errors.
+
+      /project/lib/test.ex:1:1: W: Modulewith out a @moduledoc detected
+      /project/lib/test.ex:14:1: W: Comparison to a boolean is pointless
+  """
+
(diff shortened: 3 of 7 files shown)


lib/dogma.ex
```diff
@@ -10,13 +10,33 @@ defmodule Dogma do
...
   alias Dogma.ScriptSources
 
-  def run(dir_path \\ "") do
-    dir_path
+  @formatters %{
+    "simple" => Formatter.Simple,
+    "emacs"  => Formatter.Emacs
+  }
+
+  def run(argv) do
```
lpil: Why do we want this here rather than in the mix task?

test/dogma/formatter/emacs_test.exs
```diff
@@ -0,0 +1,74 @@
...
+
+  with ".script" do
+    with "no errors" do
+      should "print nothing" do
+        script = %Script{ errors: [] }
+        assert Emacs.script(script) == ""
+      end
+    end
+
+    with "some errors" do
```
lpil: Is this correct?

lib/dogma/formatter/emacs.ex
```diff
@@ -0,0 +1,48 @@
+defmodule Dogma.Formatter.Emacs do
```
lpil: This is awesome! :)

lib/dogma.ex
```diff
@@ -10,13 +10,33 @@ defmodule Dogma do
...
+    argv
+    |> parse_args
+    |> process
+  end
+
+  def parse_args(argv) do
```
lpil: This doesn't quite feel like the right place for this to me, but I'm happy to leave it here for now as I want to overhaul this section of the application in the near future.

lib/dogma/formatter/emacs.ex
```diff
@@ -0,0 +1,48 @@
...
+  """
+  def finish(scripts) do
+    errors = scripts
+    |> format_errors
+    |> Enum.join("\n")
+
+    errors <> "\n"
+  end
+
+  def format_errors(script = %Script{}) do
```
lpil: This should be private I think?

lib/dogma/formatter/emacs.ex
```diff
@@ -0,0 +1,48 @@
...
+    |> Enum.map(&(format_error(&1, script.path)))
+  end
+
+  def format_errors(scripts) do
+    scripts
+    |> Enum.map(&format_errors/1)
+    |> List.flatten
+  end
+
+  def format_error(error, script) do
```
lpil: And this.

lpil:
This is awesome! A great step forwards. Thanks :)

lib/dogma/formatter/emacs.ex
5thWall: Honestly I was just copying the name from [`rubocop`](https://github.com/bbatsov/rubocop#formatters). I just now poked around the Syntastic checkers, and rubocop was the only one I found that called it `emacs`. For a lot of programs it seems like a similar format is just the default. Others call it `verbose` or `machine-readable`.

We could call it whatever, maybe `machine`?. We could call it `Vim` or `Syntastic` since that seems to be our intended use case. ;)

lib/dogma.ex
5thWall: I thought about pulling it out into it's own module, but it didn't seem like it would be necessary until there were more configuration options.

test/dogma/formatter/emacs_test.exs
5thWall: It's a parallel test case to https://github.com/lpil/dogma/blob/master/test/dogma/formatter/simple_test.exs#L39. In the `.script` function it doesn't seem like formatters really care about the contents of the error, just if they should note an error has happened.

lib/dogma.ex
5thWall: When I looked at the original code it felt like the mix task was mostly just delegating to the dogma run method. Though now that I think about it the mix task does seem like a better place. Do you want me to break out the option parsing into it's own module as well?

lib/dogma/formatter/emacs.ex
lpil: I *think* that the format comes from Flycheck, so I would prefer that. My second favourite is Emacs. What's your preference?

lib/dogma/formatter/emacs.ex
```diff
@@ -0,0 +1,48 @@
...
+    |> Enum.map(&(format_error(&1, script.path)))
+  end
+
+  defp format_errors(scripts) do
+    scripts
+    |> Enum.map(&format_errors/1)
+    |> List.flatten
+  end
+
+  def format_error(error, script) do
```
lpil: defp here please :)

lib/dogma/formatter/emacs.ex
5thWall: If we're not going for something general like `machine` then let's name it after Flycheck.

lpil: Great :)

I'd like to go against `machine` since I want to support other easily machine readable formats.

lib/dogma.ex
lpil: Let's leave it for now :)

lpil:
Thanks! :D

squaresurf:
How do I use this with syntastic?

PR #91, opened 2015-08-29
Add test to spot hard tabs used for indenting files

meadsteve:


lib/dogma/rules/hard_tabs.ex
```diff
@@ -0,0 +1,35 @@
...
+
+  So the following would be invalid:
+
+  def something do
+  \t:body
+  end
+  """
+  @behaviour Dogma.Rule
+
+  @indenting_tab_pattern ~r/^\t+.+/
```
lpil: Do we want to check that the line starts with a tab, or that they are in the source at all?

lib/dogma/rules/hard_tabs.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Dogma.Rules.HardTabs do
+  @moduledoc """
+  Requires that all indentation is done using spaces rather than hard tabs.
+
+  So the following would be invalid:
+
+  def something do
+  \t:body
```
lpil: Ah, I wondered how you got a tab into the documentation without triggering the linter.

lpil:
Good idea with this rule :)

lib/dogma/rules/hard_tabs.ex
meadsteve: Yeah I'm still not sure if I actually want a tab in the documentation. The fact that all blank space looks the same makes it kind of useless as documentation.  How would you feel if it read like:

```
 So the following would be invalid:

  def something do
  {tab} :body
```

lib/dogma/rules/hard_tabs.ex
meadsteve: I considered just looking for tabs in general but I thought there might be some legitimate uses cases. Looking for them at the start will spot most of the invalid cases and hopefully without any of the false positives.

lpil: Hm. I'm not sure. What's a valid use case?

lib/dogma/rules/hard_tabs.ex
lpil: I don't think I like that, but the tab is unclear too. :/

meadsteve: option two: escape the \t in the module doc so it reads something like:

```
 So the following would be invalid:

  def something do
  \t :body
```

lib/dogma/rules/hard_tabs.ex
meadsteve: Well I can't think of one (but then I don't really use hard tabs). Mainly just done it like this because the thing we want to stop is hard tabs being used to control indenting. Rather than tabs existing in the file at all.

lib/dogma/rules/hard_tabs.ex
lpil: Oh I don't know. Let's just merge this :)

lib/dogma/rules/hard_tabs.ex
lpil: Seems reasonable!

Issue #46, opened 2015-07-04
Rule: Literal in interpolation


lpil:
```elixir
"This is #{ "not" } permitted."
"Nor is #{ :this }."
```

And so on.

lpil:
@richardbatty

Might want to look at this for an idea https://github.com/lpil/dogma/blob/master/lib/dogma/rules/literal_in_condition.ex

Probably want to move the [literal check function](https://github.com/lpil/dogma/blob/master/lib/dogma/rules/literal_in_condition.ex) into a new module under `Dogma.Util`

Be sure to add the Rule behaviour, and to add the rule to the `All` set.
https://github.com/lpil/dogma/blob/master/lib/dogma/rules/sets/all.ex

PR #93, opened 2015-08-30
Add literal in interpolation rule

jacknoble:
Hopefully this is what you're looking for in https://github.com/lpil/dogma/issues/46 . Let me know if there are any test case you'd like me to add or changes you'd like made!

### DIFF
--- a/lib/dogma/rules/literal_in_interpolation.ex
+++ b/lib/dogma/rules/literal_in_interpolation.ex
@@ -1,0 +1,55 @@
+defmodule Dogma.Rules.LiteralInInterpolation do
+  @moduledoc """
+  A rule that disallows useless string interpolations that contain a literal value
+  instead of a variable or function.
+  """
+
+  @behaviour Dogma.Rule
+
+  alias Dogma.Script
+  alias Dogma.Error
+
+  def test(script, _config = [] \\ []) do
+    script |> Script.walk( &check_node(&1, &2) )
+  end
+
+  defp check_node({:::, _, [{{:., _, [Kernel, :to_string]}, meta, [interpolated_token]}, {:binary, _, _}]} = node, errors) do
+    if interpolated_token |> literal? do
+      {node, [error(meta[:line]) | errors]}
+    else
+      {node, errors}
+    end
+  end
+
+  defp check_node(node, errors) do
+    {node, errors}
+  end
+
+
+  defp literal?(x) when is_atom       x do true end
+  defp literal?(x) when is_binary     x do true end
+  defp literal?(x) when is_bitstring  x do true end
+  defp literal?(x) when is_boolean    x do true end
+  defp literal?(x) when is_function   x do true end
+  defp literal?(x) when is_list       x do true end
+  defp literal?(x) when is_map        x do true end
+  defp literal?(x) when is_pid        x do true end
+  defp literal?(x) when is_port       x do true end
+  defp literal?(x) when is_number     x do true end
+  defp literal?(x) when is_reference  x do true end
+  defp literal?({:{}, _, _}) do
+    true
+  end
+  defp literal?(_) do
+    false
+  end
+
+
+  defp error(pos) do
+    %Error{
+      rule:     __MODULE__,
+      message:  "Literal value found in interpolation",
+      line: pos,
+    }
+  end
+end
--- a/lib/dogma/rules/literal_in_interpolation.ex
+++ b/lib/dogma/rules/literal_in_interpolation.ex
@@ -9,6 +9,8 @@ defmodule Dogma.Rules.LiteralInInterpolation do
   alias Dogma.Script
   alias Dogma.Error
 
+  import Dogma.Util.ASTNode, only: [literal?: 1]
+
   def test(script, _config = [] \\ []) do
     script |> Script.walk( &check_node(&1, &2) )
   end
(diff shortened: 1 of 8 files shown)


lpil:
Hello! Thank you for contributing. Always great to get a new person involved :)

lib/dogma/rules/literal_in_interpolation.ex
```diff
@@ -0,0 +1,50 @@
...
+  alias Dogma.Script
+  alias Dogma.Error
+
+  import Dogma.Util.ASTNode, only: [literal?: 1]
+
+  def test(script, _config = [] \\ []) do
+    script |> Script.walk( &check_node(&1, &2) )
+  end
+
+  defp check_node({:::, meta, args} = node, errors) do
```
lpil: `#{}` is a call to `:::`? How strange, wouldn't have guessed that.

lib/dogma/rules/literal_in_interpolation.ex
```diff
@@ -0,0 +1,50 @@
...
+       _ -> {node, errors}
+    end
+  end
+
+  defp check_node(node, errors) do
+    {node, errors}
+  end
+
+  defp get_interpolated_token(args) do
+    case args do
```
lpil: Why do we need this error handling?

lib/dogma/util/ast_node.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Dogma.Util.ASTNode do
```
lpil: Good idea :)

test/dogma/documentation/rules_list_test.exs
```diff
@@ -9,7 +9,7 @@ defmodule Dogma.Documentation.RuleListTest do
     assert actual == expected, """
     docs/rules.md is out of date.
 
-    Run `mix run script/generate_documentation.ex` in order to regenerate
+    Run `mix run scripts/generate_documentation.exs` in order to regenerate
```
lpil: Oops, my bad.

test/dogma/util/ast_node_test.exs
```diff
@@ -0,0 +1,5 @@
+defmodule Dogma.Util.ASTNodeTest do
+  use ShouldI
+
+  doctest Dogma.Util.ASTNode
```
lpil: Probably should have more tests here.

lib/dogma/rules/literal_in_interpolation.ex
jacknoble: So one pattern Jose matches with is [here](https://github.com/elixir-lang/elixir/blob/aa901b1f5ad0749fecaf1fe5d89bbabcabf62399/lib/elixir/lib/macro.ex#L632). So `:<<>> ` is the salient function and arguments to that function are string literals and interpolations whose pattern is 
```
{:::, _,
   [{{:., _, [Kernel, :to_string]}, [], [(thing getting interpolated)]}, {:binary, _, _}]}}
```

`:::` by itself definitely over matches which is the reason for the error handling below. I originally had the whole pattern in the function def but that's pretty long and unwieldy. I'm sure you have more experience figuring out what to match on, open to suggestions.

test/dogma/util/ast_node_test.exs
jacknoble: Agreed. I'll try to add some later this evening.

lib/dogma/rules/literal_in_interpolation.ex
lpil: Lets go with this, we can refine it later if we discover problems.

jacknoble:
Added some tests for the ASTNode module. Rewrote some of the the `literal?` function so it makes more sense for AST nodes.

lib/dogma/util/ast_node.ex
```diff
@@ -0,0 +1,27 @@
...
+  def literal?(x) when is_binary  x do true end
+  def literal?(x) when is_boolean x do true end
+  def literal?(x) when is_list    x do true end
+  def literal?(x) when is_number  x do true end
+  def literal?({_,_})               do true end
+  def literal?({:{}, _, _})         do true end
+  def literal?({:%{}, _, _})        do true end
+  def literal?({:sigil_s,_,_})      do true end
+  def literal?({:sigil_S,_,_})      do true end
+  def literal?({:sigil_w,_,_})      do true end
```
lpil: What about other sigils?

test/dogma/util/ast_node_test.exs
```diff
@@ -0,0 +1,71 @@
+defmodule Dogma.Util.ASTNodeTest do
```
lpil: Looks good :)

lib/dogma/util/ast_node.ex
jacknoble: Updated to match all of them. I agree.

jacknoble:
Rebased onto current master.

lib/dogma/util/ast_node.ex
```diff
@@ -0,0 +1,31 @@
...
+  def literal?(x) when is_list    x do true end
+  def literal?(x) when is_number  x do true end
+  def literal?({_,_})               do true end
+  def literal?({:{}, _, _})         do true end
+  def literal?({:%{}, _, _})        do true end
+  def literal?(node) do
+    sigil?(node)
+  end
+
+  def sigil?({fun, _, _}) when is_atom fun do
```
lpil: If we build functions for each of these by iterating over the letters of the alphabet we will check all sigils and have no runtime penalty. Additionally, this matches function calls like this:

```elixir
def sigil_transformer(x) do
  # ...
end
```

https://github.com/lpil/dogma/blob/0d0ece43081c28059e487bd0b681068c9d3ceb4a/lib/dogma/rules/quotes_in_string.ex#L23-L26

jacknoble: Ooh, smart! Does it make sense to have macros for defining functions that match different classes of AST node? e.g.
```
def_for_sigils literal? ({sigil,_,_}) do false end
```
which would expand to define that function for all sigils?

lpil: Sure, though maybe not for now. When we have another case in which we need all sigils we should probably make this. :)

Issue #92, opened 2015-08-29
ModuleDoc should report the name of the module without a moduledoc


lpil:

lpil:
Resolved by @jacknoble !

PR #94, opened 2015-08-30
Add module name to ModuleDoc rule errors

jacknoble:
https://github.com/lpil/dogma/issues/92

### DIFF
--- a/lib/dogma/rules/module_doc.ex
+++ b/lib/dogma/rules/module_doc.ex
@@ -43,11 +43,12 @@ defmodule Dogma.Rules.ModuleDoc do
     end
   end
 
-  defp check_node({:defmodule, m, [_, [do: module_body]]} = node, errors) do
+  defp check_node({:defmodule, m, [mod, [do: module_body]]} = node, errors) do
+    {_,_,names} = mod
     if module_body |> moduledoc? do
       {node, errors}
     else
-      {node, [error( m[:line] ) | errors]}
+      {node, [error( m[:line], Enum.join(names, ".") ) | errors]}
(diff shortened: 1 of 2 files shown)


lib/dogma/rules/module_doc.ex
```diff
@@ -43,11 +43,12 @@ defmodule Dogma.Rules.ModuleDoc do
...
   end
 
-  defp check_node({:defmodule, m, [_, [do: module_body]]} = node, errors) do
+  defp check_node({:defmodule, m, [mod, [do: module_body]]} = node, errors) do
+    {_,_,names} = mod
     if module_body |> moduledoc? do
       {node, errors}
     else
-      {node, [error( m[:line] ) | errors]}
+      {node, [error( m[:line], Enum.join(names, ".") ) | errors]}
```
lpil: Can we use `Module.concat/1` please? :)

lpil:
Looks great, thank you. Just one little note above :)

lib/dogma/rules/module_doc.ex
jacknoble: Ah, good point. One thing is that
```
iex(1) [:ModName] |> Module.concat |> to_string
"Elixir.ModName"
```

That 'Elixir' prefix in the errror message seems a little noisy but if it doesn't bother you I'll change it.

lpil: Ah, I didn't think about that. What you've done is better then. :)

lpil:
Coverage dropped due to bug with detecting string interpolation.

Issue #64, opened 2015-08-06
Spike: Experiment with extracting the Elixir tokenizer from the compiler


lpil:
It could prove very handy!

lpil:
Done. Thanks to @5thWall for showing us how easy this is.

PR #97, opened 2015-08-31
Add rule disallowing semicolons

### LINKED ISSUE #47: Rule: Semicolon
Expressions should not be terminated by semicolons.

```elixir
x = 1;
y = 2;
```

5thWall:
I'm pulling in the Elixir tokenizer for this one and I know there's a spike for that #64. Do you want me to pull that into the `Dogma.Script` module? Maybe add a `Dogma.Script.tokens` field?

Closes #47

### DIFF
--- a/test/dogma/rules/semicolon_test.exs
+++ b/test/dogma/rules/semicolon_test.exs
@@ -1,0 +1,117 @@
+defmodule Dogma.Rules.SemicolonTest do
+  use ShouldI
+
+  @message "Expressions should not be terminated by semicolons."
+
+  alias Dogma.Rules.Semicolon
+  alias Dogma.Script
+  alias Dogma.Error
+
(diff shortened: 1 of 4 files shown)


lpil:
Holy crap. I had no idea that the official tokenizer was actually available in the standard library. I presumed we would have to pull those files out of the compiler project. This means we should probably refactoring a bunch of the other rules, and I need to get familiar with the tokenizer output.

Thank you! I'm gunna have to study this one for a bit before I merge it in, but it looks great so far.

test/dogma/rules/semicolon_test.exs
```diff
@@ -0,0 +1,117 @@
+defmodule Dogma.Rules.SemicolonTest do
```
lpil: Nice tests, covered a lot of cases :)

PR #102, opened 2015-09-01
New Rule: Always use `true` as the final condition of a `cond` statement

5thWall:
From the community style guide.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -1,13 +1,14 @@
 # Dogma Rules
 
 These are the rules included in Dogma by default. Currently there are
-22 of them.
+23 of them.
 
 ## Contents
 
 * [BarePipeChainStart](https://github.com/lpil/dogma/blob/master/docs/rules.md#barepipechainstart)
 * [ComparisonToBoolean](https://github.com/lpil/dogma/blob/master/docs/rules.md#comparisontoboolean)
 * [DebuggerStatement](https://github.com/lpil/dogma/blob/master/docs/rules.md#debuggerstatement)
+* [FinalCondition](https://github.com/lpil/dogma/blob/master/docs/rules.md#finalcondition)
 * [FinalNewline](https://github.com/lpil/dogma/blob/master/docs/rules.md#finalnewline)
 * [FunctionArity](https://github.com/lpil/dogma/blob/master/docs/rules.md#functionarity)
 * [FunctionName](https://github.com/lpil/dogma/blob/master/docs/rules.md#functionname)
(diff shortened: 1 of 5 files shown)


lib/dogma/rules/final_condition.ex
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  defp check_node(node, errors, _) do
+    {node, errors}
+  end
+
+  defp find_last_cond([[do: children] | _]) do
+    children
+    |> Enum.reverse
+    |> hd
```
jacknoble: List.last/1 ?

lpil:
I like that we have this rule, but I'm not sure I like using true as the catch all. It doesn't read overly well, and it's not in keeping with the Elixir standard of using `_` as the catch all.

meadsteve:
Interestingly this community style guide recommends using true as the catch all http://elixir.community/styleguide however I agree with @lpil and think it doesn't read very well.

lpil:
`_` works as a catch all in a `cond`, correct? If so, I think we should recommend that as it is the same as `case` and pattern matching function definitions.

I am also rather partial to `:otherwise`, though that's probably the Haskell experience talking.

lib/dogma/rules/final_condition.ex
lpil: Thanks :)

meadsteve:
_ definitely works. And it's what I'd prefer but the defaults for this project should really be the accepted community standard (if one exists). In this case I'm not actually sure whether it's ```true``` or ```_```

lpil:
Looking at elixir-lang/ex_doc it seems to be `true`.

What should this rule do when we explicitly do not want a catch all?

5thWall:
`_` is the catch all for pattern matching, but there's nothing to match against in a `cond` statement. It doesn't really make sense to me why it works at all, except just to match `case` statements.

For when we don't want a catch all, I'm actually not sure. The community styleguide just says that the final condition should always be true. `cond` statements will fail at runtime if at least one statement doesn't evaluate to truthy. So this could guard against that, but maybe someone want's to fail fast.

lpil:
> cond statements will fail at runtime if at least one statement doesn't evaluate to truthy. So this could guard against that, but maybe someone want's to fail fast.

That's kind of the aim of the game with Elixir and Erlang in my opinion!

5thWall:
True. I could set it up so that it checks to see if the final condition is a literal value, and only returns an error then. That way it would skip over statements where the final condition is a method call or other expression.

meadsteve:
At the risk of overcomplicating our config we could make it an option whether a catch-all is required or not.

lpil:
I disagree, it should be considered on a case by case basis.

Checking checking if it is a literal or not sounds good. We do have a module that checks whether an AST node is a literal or not.

5thWall:
Okay, I'll add checking to see if the last value is a literal and compare there. Where are we on what the default should be?

lpil:
I vote `true`, as that seems to be what Jose uses.

docs/rules.md
```diff
@@ -75,6 +76,67 @@ A rule that disallows calls to IEx.pry, as while useful, we probably don't
...
+      1 + 2 == 5 ->
+        "Nope"
+      1 + 3 == 5 ->
+        "Nada"
+      _ ->
+        "OK"
+    end
+
+This rule will only catch those `cond` statements where the last condition
+is a literall or a `_`. Complex expressions and function calls will not
```
lpil: Typo, 'literal'

lpil:
Nice work. Happy to merge after the typo fix.

PR #106, opened 2015-09-09
Add current value to error messages

rrrene:
This adds the current value (which is violating some rule) to error message for `LineLength` and `FunctionArity`.

This seems like a great project! :+1:

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -200,7 +200,8 @@ So the following would be invalid:
 
 ### LineLength
 
-A rule that disallows lines longer than 80 characters in length.
+A rule that disallows lines longer than X characters in length (defaults to
+80).
 
 This rule can be configured with the `max_length` option, which allows you to
 specify your own line max character count.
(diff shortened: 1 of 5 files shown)


meadsteve:
@rrrene looks good. But the tests will need updating with the new expected messages.

rrrene:
Ah, sorry. I force pushed the corrected version!

meadsteve:
:+1:

lpil:
I actually removed this from some rules as I found people kept mistaking it for line numbers. I think your message is better than mine though.

I don't like this defaults to nil stuff in the docs. It doesn't seem like useful information to the user.

lpil:
+ really cool to see you contributing, thank you! :D

rrrene:
Jesus, I'm really on fire with stupid mistakes tonight. Of course, that should read "defaults to 80", as that is what is (a) correct and (b) actually useful information for the user.

Pushed a correction. I really need to get some sleep.

lpil:
Looks great, thank you very much. :)

PR #107, opened 2015-09-10
evalutate -> evaluate

knewter:


lpil:
Thanks! Could you please get the tests green? It's failing because the documentation needs to be regenerated.

knewter:
Did that and force-pushed over the previous commit to avoid unnecessary commits in the history

lpil:
Great! Thank you :)

PR #110, opened 2015-09-12
mix.exs: add lib/dogma.ex

henrik:
Getting errors due to this missing.

```
$ mix deps.get
…
$ mix dogma
…
** (UndefinedFunctionError) undefined function: Dogma.run/1 (module Dogma is not available)
    Dogma.run({nil, Dogma.Formatter.Simple})
    lib/mix/tasks/dogma.ex:12: Mix.Tasks.Dogma.run/1
    (mix) lib/mix/cli.ex:55: Mix.CLI.run_task/2
    (elixir) lib/code.ex:307: Code.require_file/2
```

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -35,6 +35,7 @@ defmodule Dogma.Mixfile do
     ~w(
       LICENCE
       README.md
+      lib/dogma.ex
       lib/dogma
       lib/mix/tasks/dogma.ex
       mix.exs


henrik:
Or actually, I would suggest just removing the explicit file list – the default value should cover this fine, right? https://hex.pm/docs/publish#example-mix-exs-file

lpil:
Oh christ, how long has this been broken? :(

The thing we added the file list for no longer exists, so I'll remove it. Thanks for your report!

Issue #25, opened 2015-06-21
Optionally auto fix violations where possible + safe


errm:
Rewrite to correct the failing rules where possible. As in go fmt or rubocop -a

lpil:
Certainly needed. Sounds hard.

lpil:
Does anyone have any ideas as to how we might do this? Sounds difficult. Maybe we should look into how Rubocop does it.

meadsteve:
What rules do we currently have that we'd want to consider doing this on? The ones that comes to mind at the moment are the WindowsLineEnding and TrailingWhiteSpace.

lpil:
They sound like easy wins, and I think we might be able to manage some others, but I don't really know what's the best way to architect this bit of the system.

meadsteve:
First thoughts would be. Have a Fixable behaviour then do the following:

1. Run all tests.
2. Prompt if any of the errors are fixable.
3. Apply fixes.
4 Rerun tests that are fixable and output updated number of violations.

For simple rules like the ones I mentioned the fixing could be as simple as running a map over each source line. It'd definitely be harder for rules that make use of the ast.

lpil:
I mean internally, rather than from the user interaction point of view. Will give it some thought.

meadsteve:
Internally I'd be tempted to handle this with a function that takes all the lines of a script and returns all the lines of the "fixed" file. I think this would work for all the simple ones which I think would be the only ones we'd want  to automatically fix.

lpil:
I don't want to re-run the test correction loop as it's possible to get into an infinite loop here.

meadsteve:
agreed. I don't seen any reason why you'd run any fixing steps a 2nd time. If any errors remain after automatic fixes then it's probably something that can't be fixed automatically.

PR #114, opened 2015-09-12
Fix violations

jacknoble:
Thought I'd take a crack at https://github.com/lpil/dogma/issues/25

Allows a user to pass --fix to the mix task and trailing whitespace errors will get fixed.
Other rules can do the same by implementing a correction callback.
Couple things I could use guidance on:
- Is there a way to officially define an optional callback as part of a behaviour? 
- Is this how we want to handle formatting for fixes?
- Obviously this whole approach could be bad for some reason.

### DIFF
--- a/lib/dogma/rules/trailing_whitespace.ex
+++ b/lib/dogma/rules/trailing_whitespace.ex
@@ -13,6 +13,13 @@ defmodule Dogma.Rules.TrailingWhitespace do
     Enum.reduce( script.processed_lines, [], &check_line(&1, &2) )
   end
 
+  def correction(script, error) do
+    script.source
+    |> String.split("\n")
+    |> Enum.map(&String.rstrip/1)
+    |> Enum.join("\n")
+  end
+
   defp check_line({i, line}, errors) do
     trimmed_line = String.replace(line, ~r/\r\z/, "")
     case Regex.run( @regex, trimmed_line, return: :index ) do
--- a/lib/dogma/rules/trailing_whitespace.ex
+++ b/lib/dogma/rules/trailing_whitespace.ex
@@ -13,11 +13,21 @@ defmodule Dogma.Rules.TrailingWhitespace do
     Enum.reduce( script.processed_lines, [], &check_line(&1, &2) )
   end
 
-  def correction(script, error) do
-    script.source
-    |> String.split("\n")
-    |> Enum.map(&String.rstrip/1)
-    |> Enum.join("\n")
+  def correction(source, errors) do
+    lines = Enum.map(errors, fn err -> err.line end)
+    [newline] = Regex.run(~r/\r?\n\z/, source)
+
+    source
+    |> String.split(newline)
+    |> Enum.with_index
+    |> Enum.map(fn {line, i} ->
+      if Enum.member?(lines, i + 1) do
+        String.replace(line, @regex, "")
+      else
+        line
+      end
+    end)
+    |> Enum.join(newline)
   end
 
   defp check_line({i, line}, errors) do
--- a/lib/dogma/script.ex
+++ b/lib/dogma/script.ex
@@ -121,8 +121,9 @@ defmodule Dogma.Script do
 
   @doc """
   Builds the corrected_source by running the corrections
-  for any rule that has implemented correction/2. Then replaces the file at
-  at the path with the corrected source. The correction is run once for each rule.
+  for any rule that has implemented correction/2.
+  Then replaces the file at at the path with the corrected source.
+  The correction is run once for each rule.
   """
   def repair(script, io \\ File) do
     script.errors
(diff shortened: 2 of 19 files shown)


lib/dogma/rules/trailing_whitespace.ex
```diff
@@ -13,6 +13,23 @@ defmodule Dogma.Rules.TrailingWhitespace do
...
+    |> String.split(newline)
+    |> Enum.with_index
+    |> Enum.map(fn {line, i} ->
+      if Enum.member?(lines, i + 1) do
+        String.replace(line, @regex, "")
+      else
+        line
+      end
+    end)
+    |> Enum.join(newline)
```
lpil: Can't we replace all of this with one regex? Something like this maybe:

```elixir
String.replace( str, ~r/\s+\n/, "\n" )
```

Both of these solutions strip trailing whitespace from strings, which is behaviour we don't want.

lib/dogma/script.ex
```diff
@@ -118,6 +119,44 @@ defmodule Dogma.Script do
...
+  Builds the corrected_source by running the corrections
+  for any rule that has implemented correction/2.
+  Then replaces the file at at the path with the corrected source.
+  The correction is run once for each rule.
+  """
+  def repair(script, io \\ File) do
+    script.errors
+    |> Enum.group_by(&Map.get(&1, :rule))
+    |> Enum.reduce(script, &repair_violation/2)
+    |> write_corrected_source(io)
```
lpil: I don't think that the script should be responsible for doing any IO.

lib/dogma/script.ex
```diff
@@ -118,6 +119,44 @@ defmodule Dogma.Script do
...
+  defp repair_violation({rule, violations}, script) do
+    if function_exported?(rule, :correction, 2) do
+      source = script.corrected_source || script.source
+      errors = Enum.map(script.errors, fn err ->
+        if Enum.member?(violations, err) do
+          %Error{ err | fixed?: true}
+        else
+          err
+        end
+      end)
```
lpil: I think this function is hard to read. Perhaps repairing should have its own module, and this one should delegate to it.

lib/dogma/rules/trailing_whitespace.ex
jacknoble: By only fixing lines where the error was detected we avoid stripping strings -- unless there is an edge case where that isn't true.

I think you're right that this can be simplified with a better regex, though.

lpil: Ah, right, of course, my bad.

We could just use the lines already in the struct, rather than creating them again.

lpil:
I'm not really sure about this. I don't have a clear idea of how we ought to architect this system.

Did you read how other linters do this? It would be good to see what they do.

jacknoble:
So RuboCop does this: 
1. Cops which support corrections define an #autocorrect method. This takes a node and returns a "correction" which is any object which responds to #call. Typically lambdas which have the source_buffer in scope. E.g. https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/trailing_whitespace.rb#L22 removes the characters in the ranger the cop detected the trailing whitespace.

2. As offenses are detected and collected each cop stores its corrections. https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/cop.rb#L154

3. For each file with an error, with corrections *one* cop's corrections will be run. https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/team.rb#L61 and the source buffer rewritten with that cop's changes. https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/team.rb#L77

4. Each file is looped over like this until the source doesn't change.
https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/runner.rb#L123 (this includes infinite loop detection, they check to make sure the same source isn't produced twice).

Recursing over the source seems like it might be inevitable especially once fixes start changing line counts. But we might be able to get pretty far with idempotent corrections that don't introduce new errors. There also my be an idempotent order in which changes can be applies i.e. always remove extra empty lines after removing trailing whitespace. I don't think design decisions made now will make changing to the recursive approach much harder or easier.

We could build the corrections when the rules are detecting offenses as RuboCop does and store them as closures in %Error{}. I'm not sure I see a big advantage to doing that unless the corrections are duplicating expensive work. It seems slightly less straightforward to me.

I don't know. It's definitely a hard problem.

lib/dogma/rules/trailing_whitespace.ex
jacknoble: Hmm. When there are multiple rules with corrections we'll be transforming the corrected source instead of or the original so the existing lines won't work. But if we want to run them recursively and only make one set of corrections per run that would work.

jacknoble:
Sorry I haven't had time to work on this. Will have a new version that's much closer to Rubocop's implementation this weekend.

lpil:
No worries at all! Currently I'm leaning towards holding off on this until we've got configuration sorted, some more rules made, and a bunch of bugs squished.

jacknoble:
Reasonable, I'll work on something else for now.

PR #115, opened 2015-09-13
Add failing test for BarePipeChainStart

rrrene:
It should not fail when we start with a simple lookup on a keyword list
or map. Currently this fails:

    map[:foo]
    |> is_atom
    |> IO.inspect

which would have to be written like this to pass:

    map
    |> Keyword.get(:foo)
    |> is_atom
    |> IO.inspect

I was not yet able to fix this, but will try to provide a fix later this week. For now, a failing test should remind us that this rule needs some love :wink:


lpil:
Thanks!

PR #116, opened 2015-09-17
Add new Rule: ExceptionName

rrrene:
This PR adds a rule that checks that all exception module are named with the suffix `Error` ([elixir_style_guide reference](https://github.com/niftyn8/elixir_style_guide#exceptions))

For example, prefer this:

      defmodule BadHTTPCodeError do
        defexception [:message]
      end

Not one of these:

      defmodule BadHTTPCode do
        defexception [:message]
      end

      defmodule BadHTTPCodeException do
        defexception [:message]
      end

I think in the long run we will want to be able to configure which term is used as unified suffix, but for now this helps us achieve better "coverage" of the Elixir Styleguide by niftyn8.

### DIFF
--- a/lib/dogma/rule/exception_name.ex
+++ b/lib/dogma/rule/exception_name.ex
@@ -1,0 +1,67 @@
+defmodule Dogma.Rule.ExceptionName do
+  @moduledoc """
+  A Rule that checks that exception names end with a trailing Error.
+
+  For example, prefer this:
+
+      defmodule BadHTTPCodeError do
+        defexception [:message]
+      end
+
+  Not one of these:
+
+      defmodule BadHTTPCode do
+        defexception [:message]
+      end
+
+      defmodule BadHTTPCodeException do
+        defexception [:message]
+      end
+  """
+
+  @behaviour Dogma.Rule
+  @good_name_suffix "Error"
+
+  alias Dogma.Script
+  alias Dogma.Error
+
+  def test(script, _config = [] \\ []) do
+    script |> Script.walk( &check_node(&1, &2) )
+  end
+
+  defp check_node(
+    {:defmodule, _metadata, [{:__aliases__, _, _name_arr}, children]} = node,
+    errors
+  ) do
+    error_module? =
+      children[:do]
+      |> block_children_as_list
+      |> Enum.any?(&exception?/1)
+    name = Enum.join(_name_arr, ".")
+
+    if error_module? && bad_name?(name) do
+      {node, [error(_metadata[:line], name) | errors]}
+    else
+      {node, errors}
+    end
+  end
+  defp check_node(node, errors) do
+    {node, errors}
+  end
+
+  defp bad_name?(name), do: !String.ends_with?(name, @good_name_suffix)
+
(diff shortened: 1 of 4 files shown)


lpil:
I really like this rule! :)

Got some long lines here. How many characters is that error message going to be with normal exception names? I'd like to keep them under 80 - line number characters, to keep the output neat on all terminals.

rrrene:
*Re: long lines* 

Yeah, I saw that and decided to submit the PR anyway since this touches on something I wanted to discuss: with sufficiently deep pattern matching, there will always be cases where this rule is broken. However, I am no fan of breaking pattern matches up into multiple lines (or at least I am no fan of forcing it). 

The question is: should we consider a configuration option to skip the rule for such scenarios (would be happy to work on this), so could make it optional for people like me to say "I want to adhere to the 80 char limit, except for `def`, `defp`, et al."?

*Re: output length* 

I see why you are concerned about this. I am not sure what qualifies as "normal exception name", maybe I will do a quick grep across some popular repos ...

lpil:
I think that if your pattern matching is that deep, you probably want to break it into multiple functions or lines in order to make it understandable. Pattern matching on 2 nested function calls is probably as much as I'd be comfortable with.

```elixir
  defp check_node({:defmodule, _metadata, [{:__aliases__, _, _name_arr}, children]} = node, errors) do
```

```elixir
  defp check_node(
    {:defmodule, _, [{:__aliases__, _, _}, [do: body]]} = node,
    errors
  ) do
```

rrrene:
Good call. I will incorporate your suggestions this weekend and update the PR!

rrrene:
I updated this PR. The failing test coverage test on this PR is not about code I touched, so I think we're good?

lib/dogma/rule/exception_name.ex
```diff
@@ -0,0 +1,68 @@
...
+  def block_children_as_list({:__block__, _, children}), do: children
+  def block_children_as_list(children), do: [children]
+
+  defp exception?({:defexception, _, _}), do: true
+  defp exception?(_), do: false
+
+  defp error(line, _name) do
+    %Error{
+      rule:     __MODULE__,
+      message:  "Exception names should end with a trailing " <>
```
lpil: I wonder if the compiler optimises this binary concat away.

test/dogma/rule/exception_name_test.exs
```diff
@@ -0,0 +1,44 @@
...
+      defexception [:message]
+    end
+    defmodule BadHTTPCodeException do
+      defexception [:message]
+    end
+    """ |> lint
+    expected_errors = [
+      %Error{
+        rule:     ExceptionName,
+        message:  "Exception names should end with a trailing 'Error'.",
```
lpil: Sorry, but can we change this to "Exception names should end with 'Error'." please? :)

lpil:
Looks great, thanks! One little note above.

RE coverage- there's a slight bug with how the coverage tool picks up string interpolation, it's a bit annoying.

lpil:
Thank you!

PR #118, opened 2015-09-20
Clean up code in Dogma.Util.ASTNode

rrrene:
This cleans up some code to adhere to the style guide we are sometimes
referring to.

So this is just housekeeping.


lpil:
Thanks :)

Issue #117, opened 2015-09-20
Why are there functions called probably_X?


rrrene:
Is there a reason why the naming utility functions are named and coded so defensively?

https://github.com/lpil/dogma/blob/91981ea1b2a44b3f3660d88c172f099f507d86bd/lib/dogma/util/name.ex#L15

There might be a good reason for this, but seeing it I'm just thinking "can't we just ensure it's a snake_case name and call it that?"

lpil:
Because it doesn't really check if something is snake case or not, because I don't really know how I'd check that. Probably should make the name more sensible now.

rrrene:
I could refactor this into `snake_case?` and so on, I was just curious if you had any special reasons.

We should be able to check with certainty :wink:

lpil:
Go for it :)

rrrene:
Done. Closing this ticket in favor of the PR.

PR #119, opened 2015-09-20
Replace probably_* functions with decisivei ones

rrrene:
So probably_snake_case? becomes snake_case? and so on ...

Discussed here: https://github.com/lpil/dogma/issues/117

### DIFF
--- a/lib/dogma/documentation/formatters_list.ex
+++ b/lib/dogma/documentation/formatters_list.ex
@@ -5,7 +5,7 @@ defmodule Dogma.Documentation.FormattersList do
 
   alias Dogma.Formatter
 
-  @doc"""
+  @doc """
   Writes formatter documentation to `docs/formatters.md`.
   """
   def write! do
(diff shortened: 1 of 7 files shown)


lpil:
Thanks! :)

lpil:
Dogma seems to be failing now.

```
Inspecting 97 files.

.........................................................X.......................................

97 files, 1 error!

== test/dogma/script_sources_test.exs ==
54: VariableName: Variable names should be in snake_case
```

rrrene:
Yeah, it was, but not because of the `snake_case?` test. It was failing because the VariableName rule was not working correctly. 

Add this code to a test using the current master and it will pass:

    [
      %ScriptSource{some_field: someVariable}
    ] = anotherVar

The functionality in the Name module works correctly and I tried my best to fix the rule for now, but discovered deeper issues with the VariableName rule in the process. 

The most prominent one is that we stop to discover errors in an AST-Node when we find the first one (due to the use of `Enum.all?`). We also get wrong line numbers with the sample code above.

Maybe I will try to find a solution for this next week, but at least this PR should now be fixed.

lpil:
Thanks :)

Did you open an issue for this bug?

rrrene:
I opened an issue for it!

lpil:
Thanks

PR #121, opened 2015-09-23
Fix BarePipeChainStart for ranges

### LINKED ISSUE #120: BarePipeChainStart incorrectly fails for ranges
```elixir
0..14 |> Enum.map(fn n -> n * 2 end)
```

rrrene:
Fixes #120 (also adds a test to ensure it stays fixed ^_^)

### DIFF
--- a/lib/dogma/rule/bare_pipe_chain_start.ex
+++ b/lib/dogma/rule/bare_pipe_chain_start.ex
@@ -80,6 +80,10 @@ defmodule Dogma.Rule.BarePipeChainStart do
   defp function_line({:<<>>, _, _}),
   do: :ok
 
+  # exception for binaries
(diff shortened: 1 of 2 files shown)


lib/dogma/rule/bare_pipe_chain_start.ex
```diff
@@ -80,6 +80,10 @@ defmodule Dogma.Rule.BarePipeChainStart do
   defp function_line({:<<>>, _, _}),
   do: :ok
 
+  # exception for binaries
```
lpil: For binaries? :)

rrrene: Ahem ... yes? no? :sweat_smile:

lpil:
Thanks! :)

Issue #123, opened 2015-09-24
Formatters fail if Dogma is a dependency


rrrene:
I tried to add Dogma to another project (as a dependency) and get the following when I run the Mix task:

    $ mix dogma --format=json
    ** (UndefinedFunctionError) undefined function: Dogma.Mixfile.project/0 (module Dogma.Mixfile is not available)
        Dogma.Mixfile.project()
        lib/dogma/formatter/json.ex:70: Dogma.Formatter.JSON.metadata/0
        lib/dogma/formatter/json.ex:54: Dogma.Formatter.JSON.finish/1
        lib/dogma/formatter.ex:49: Dogma.Formatter.finish/2
        lib/mix/tasks/dogma.ex:12: Mix.Tasks.Dogma.run/1
        (mix) lib/mix/cli.ex:55: Mix.CLI.run_task/2

The problem: `Dogma.Mixfile.project/0` and therefore the `dogma` Mix task is only available within the Dogma project itself. For Dogma to be interoperable within other projects, this should not be the case.

lpil:
Is this just the JSON formatter? It looks like it is, going by the stack trace.

PR #124, opened 2015-09-24
Add Dogma.version

rrrene:
This fixes the issue raised in #123 while keeping the version number in the JSON output **and** providing the handy `Dogma.version` function for future use.

This bug only came up because I wanted to use Dogma in a test project, where I wanted to create a custom Rule for "method length in lines" ... If you have any interests in exploring this topic of "Dogma as a dependency" further, I could open source my "wrapper code" on the weekend!

### DIFF
--- a/lib/dogma.ex
+++ b/lib/dogma.ex
@@ -10,6 +10,8 @@ defmodule Dogma do
   alias Dogma.Rules
   alias Dogma.ScriptSources
 
+  @version Mix.Project.config[:version]
+
   def run({dir, formatter}) do
     dir
     |> ScriptSources.find(exclude_patterns)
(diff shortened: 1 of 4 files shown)


lib/dogma.ex
```diff
@@ -10,6 +10,8 @@ defmodule Dogma do
   alias Dogma.Rules
   alias Dogma.ScriptSources
 
+  @version Mix.Project.config[:version]
```
lpil: Won't this return the version of the project that requires Dogma, not Dogma itself?

rrrene: Nope. It's read at *compile* time for the individual project. I'm using the same "trick" in InchEx's Formatter to add the InchEx version to JSON output: https://github.com/rrrene/inch_ex/blob/99987b500d5f14033645a68efb4fde7f7aa11a7b/lib/inch_ex/docs/formatter.ex#L2

lpil: Ah, very nice. :)

PR #125, opened 2015-09-26
Test more elixir versions using travis

meadsteve:
Including 1.1.0 & otp 18

PR #126, opened 2015-09-26
Remove underscores from used variables in exception rule

meadsteve:
Found this mistakes when running #125 against elixir 1.1.0.


lpil:
Huh, those shouldn't have been underscored. Good change.

PR #127, opened 2015-09-26
Update travis test dependencies

timbuchwaldt:


timbuchwaldt:
So - this works for Elixir 1.1.0 now  (although a bit hacky). Do you want to still support 1.0.x, @lpil ? 
I would otherwise refactor this a bit. Supporting 1.0.x probably needs some more magic pattern matching to detect the new output of tokenize.

timbuchwaldt:
Haha nice. Should have checked the other PRs first m(

lpil:
Oh dear. I guess this is what we get for making bits of the compiler hard dependencies... ;)

I've not had time to look into this, but we'll find a resolution soon. I'd like to maintain compatibility with all supported versions of Elixir.

lpil:
@rrrene fixed it before I even got to a computer! :dancer: 

Could you rebase please?

PR #139, opened 2015-10-09
rule for interpolation-only strings

alxndr:
Rule to favor `foo(bar)` over `"#{foo(bar)}"`.

Not sure if this covers nested strings or destructuring or other interesting/edge cases...

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -1,7 +1,7 @@
 # Dogma Rules
 
 These are the rules included in Dogma by default. Currently there are
-26 of them.
+27 of them.
 
 ## Contents
 
@@ -23,6 +23,7 @@ These are the rules included in Dogma by default. Currently there are
 * [ModuleName](#modulename)
 * [NegatedAssert](#negatedassert)
 * [NegatedIfUnless](#negatedifunless)
+* [NoInterpolationOnlyStrings](#nointerpolationonlystrings)
 * [PredicateName](#predicatename)
 * [QuotesInString](#quotesinstring)
 * [Semicolon](#semicolon)
@@ -361,6 +362,18 @@ These are considered invalid:
     unless not sad? do
       mope_about()
     end
+
+
+### NoInterpolationOnlyStrings
+
+A rule that disallows strings which are entirely the result of an
+interpolation.
+
+Good:
+    output = "Dogma.Rule.NoInterpolationOnlyStrings"
+
+Bad:
+    output = inspect(__MODULE__)
 
 
 ### PredicateName
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -370,10 +370,12 @@ A rule that disallows strings which are entirely the result of an
 interpolation.
 
 Good:
-  output = inspect(__MODULE__)
+
+      output = inspect(self)
 
 Bad:
-  output = "Dogma.Rule.NoInterpolationOnlyStrings"
+
+      output = "#{inspect self}"
 
 
 ### PredicateName
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -370,10 +370,10 @@ A rule that disallows strings which are entirely the result of an
 interpolation.
 
 Good:
-    output = "Dogma.Rule.NoInterpolationOnlyStrings"
+  output = inspect(__MODULE__)
 
 Bad:
-    output = inspect(__MODULE__)
+  output = "Dogma.Rule.NoInterpolationOnlyStrings"
 
 
 ### PredicateName
--- a/test/dogma/rule/no_interpolation_only_strings_test.exs
+++ b/test/dogma/rule/no_interpolation_only_strings_test.exs
@@ -35,7 +35,7 @@ defmodule Dogma.Rule.NoInterpolationOnlyStringsTest do
   should "not error for a string which includes more than an interpolation" do
     errors = ~S"""
     who = "world"
-    Hello #{who}
+    "Hello #{who}"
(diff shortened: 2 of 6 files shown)


docs/rules.md
```diff
@@ -363,6 +364,18 @@ These are considered invalid:
     end
 
 
+### NoInterpolationOnlyStrings
+
+A rule that disallows strings which are entirely the result of an
+interpolation.
+
+Good:
+    output = "Dogma.Rule.NoInterpolationOnlyStrings"
```
alxndr: Looks like part of the moduledoc is being interpreted... Expected? Should I just pick a more innocuous thing to inspect in my example?

lpil: The module doc currently breaks the rule. Make it a ~S""" sigil to prevent interpolation. :)

Also, I think the good and bad examples are switched.

lpil:
Looks good, but it needs tests before we can merge it in. :)

docs/rules.md
alxndr: If I switch it to a `~S"""` string, then it won't be an example of a bad string any more, will it?

alxndr:
Ah right you are! I'm using the `QuotesInStringTest` as inspiration, and have it mostly working, but I'm not quite sure how the testing is expected to work... In [this test case](https://github.com/lpil/dogma/blob/001245243b90ad086c00d9d5c1e0ec098cf85114/test/dogma/rule/no_interpolation_only_strings_test.exs#L35-L41) the test code is being eval'd in a context which doesn't understand creating a variable:
```
** (CompileError) test/dogma/rule/no_interpolation_only_strings_test.exs:38: function who/0 undefined
    (stdlib) lists.erl:1336: :lists.foreach/2
    (stdlib) erl_eval.erl:657: :erl_eval.do_apply/6
```
Any advice? Is there a better way to test this?

docs/rules.md
```diff
@@ -363,6 +364,18 @@ These are considered invalid:
...
+### NoInterpolationOnlyStrings
+
+A rule that disallows strings which are entirely the result of an
+interpolation.
+
+Good:
+  output = inspect(__MODULE__)
+
+Bad:
+  output = "Dogma.Rule.NoInterpolationOnlyStrings"
```
alxndr: TODO: how to write something like `"#{inspect var}"` in the `@moduledoc` without having `var` be interpreted

docs/rules.md
lpil: For the string that is the argument to the moduledoc, not for the string in the example. Currently the #{} is being interpolated into the moduledoc's string, and thus doesn't show in the rendered example.

lpil:
RE this -> https://github.com/lpil/dogma/blob/001245243b90ad086c00d9d5c1e0ec098cf85114/test/dogma/rule/no_interpolation_only_strings_test.exs#L35-L41

This is the same problem as the moduledoc above. You want your string to contain  `#{foo}`, but if you use a regular string you have to escape it, otherwise the interpolation happens in the string containing the code example. You can do this by typing `#\{foo}` (note the slash before), or by changing the string containing the example code to a `~S` sigil.

See here

https://github.com/lpil/dogma/blob/master/test/dogma/rule/literal_in_interpolation_test.exs
https://github.com/lpil/dogma/blob/master/lib/dogma/rule/literal_in_interpolation.ex

alxndr:
Ah, I see what you're saying, thanks.

alxndr:
Not sure what to do about the Coveralls failure.

lpil:
I think it's a bug in coveralls. I'll review this when I'm at a computer. :)

test/dogma/rule/quotes_in_string_test.exs
```diff
@@ -54,7 +54,7 @@ defmodule Dogma.Rule.QuotesInStringTest do
 
   should "not error for a quote in a ~R regex" do
     errors = """
-    ~r/"/
+    ~R/"/
```
lpil: Oops. Thanks!

test/dogma/rule/no_interpolation_only_strings_test.exs
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  should "not error for a quote in a ~S string" do
+    errors = """
+    ~S(hello, quote -> " <-)
+    """ |> lint
+    assert [] == errors
+  end
+
+  should "not error for a quote in a heredoc" do
```
lpil: These quote tests have nothing to with interpolation, can we remove them please? :)

test/dogma/rule/no_interpolation_only_strings_test.exs
```diff
@@ -0,0 +1,85 @@
...
+    errors = """
+    "Hello, world!"
+    """ |> lint
+    assert [] == errors
+  end
+
+  should "not error for a string which includes more than an interpolation" do
+    errors = ~S"""
+    who = "world"
+    Hello #{who}
```
lpil: This isn't valid Elixir syntax.

test/dogma/rule/no_interpolation_only_strings_test.exs
alxndr: Sure thing

test/dogma/rule/no_interpolation_only_strings_test.exs
alxndr: Whoops! Thanks for catching that.

lpil:
Thanks!

PR #140, opened 2015-10-09
test coverage for ignoring escaped things in docstrings

alxndr:
Try as I might I couldn't get the "escaped doublequote in normal string" case... Would have thought that this would catch it:

    should "strip escaped doublequote in strings" do
      processed = ~S"""
      defmodule Quotes do
        def doublequote() do
          "here is a quotation mark: \\\"."
        end
      end
      """ |> ScriptStrings.blank
      desired = """
      defmodule Quotes do
        def doublequote() do
          ""
        end
      end
      """
      assert processed == desired
    end


mix.exs
```diff
@@ -33,7 +33,7 @@ defmodule Dogma.Mixfile do
   defp deps do
     [
       # Test framework
-      {:shouldi, only: :test},
+      {:shouldi, only: ~w(dev test)a},
```
lpil: Why is it in dev now?

alxndr: It looks like Coveralls is meant to be available in the dev environment, but it depends on ShouldI. Adding ShouldI to dev means that Coveralls can be run in dev too.

lpil: Run coveralls in test rather than pollute dev with test deps. :)

alxndr: Ok, should I take excoveralls out of the dev environment then as well?

lpil: Probably a good idea!

alxndr: You got it!

Do you think it's worth mentioning the coveralls task in the readme, perhaps [in the contributor section](https://github.com/lpil/dogma/tree/82d7e06dbdcab444c1a781ca796f1e1f5bb222a0#test-it)?

lpil: Possibly. I don't think it matters so much as it's running on CI.

README.md
```diff
@@ -88,11 +88,11 @@ mix docs
 ```
 
 
-# LICENCE
+# LICENSE
 
 ```
 Dogma - A code style linter for Elixir, powered by shame.
-Copyright © 2015 Louis Pilfold - MIT Licence
+Copyright © 2015 Louis Pilfold - MIT License
```
lpil: Licen***c***e please.

meadsteve: @lpil you really should follow that comment with a giant British flag :smile:

alxndr: Whoops, sorry! That's a regional difference which I always forget about.

lpil: :gb:

meadsteve: :tea:

lpil:
Happy to merge?

PR #141, opened 2015-10-11
Rules module tests

### LINKED ISSUE #137: /lib/dogma/rules.ex has no tests

meadsteve:
Fixes #137 

This is still a bit of a work in progress. I've added mock so that we can test that Dogma.RuleSet.All is used correctly but there are other options.

Thoughts?


lpil:
I don't like that we're using a mocking library. Also, tests cannot be run async when using meck.

meadsteve:
@lpil I'm not super keen on mocks. And we definitely need to limit their usage. But it's a nice way of testing that the default set gets loaded as expected. 

I've added an additional test that uses rule set provided by setting config. If mocking is deal breaker I can always remove that test.

test/dogma/rules_test.exs
```diff
@@ -0,0 +1,79 @@
...
+
+    AssertingFormatter.start_listening
+    Rules.test([single_script], AssertingFormatter)
+
+    AssertingFormatter.assert_script_called_with script_after_run
+  end
+
+end
+
+defmodule Dogma.RuleSets.FakeRuleSet do
```
lpil: Can we move this module inside the module above please? Just to reduce the chance of accidentally having a namespace collision in other tests.

test/dogma/rules_test.exs
```diff
@@ -0,0 +1,79 @@
...
+  def test(_script, _config = [] \\ []) do
+    [:always_fake_error]
+  end
+end
+
+defmodule Dogma.Test.AssertingFormatter do
+  import ExUnit.Assertions
+
+  def start_listening do
+    Process.register self, Dogma.Test.AssertingFormatter
```
lpil: `__module__`

test/dogma/rules_test.exs
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  def script(script, _formatter \\ nil) do
+    send Dogma.Test.AssertingFormatter, {:script, script}
```
lpil: Why do we need this at all? Can't we just put the assertion in this function?

lpil:
I would rather drop the tests using meck for now.

test/dogma/rules_test.exs
meadsteve: We can but we can't move the ```Dogma.Rule.FakeRule``` module because the rules always get prefixed with ```Dogma.Rule.```.

test/dogma/rules_test.exs
meadsteve: good call.

test/dogma/rules_test.exs
meadsteve: It seemed a bit odd putting the assertion this far away from the actual test. Though I guess the name of the module may be enough.

test/dogma/rules_test.exs
lpil: `Dogma.Rule.RulesTest.FakeRule` ?

meadsteve:
@lpil cool. I'll drop meck/mock for now. We've got sufficient coverage without it.

lpil:
Thank you. :)

test/dogma/rules_test.exs
meadsteve: but that won't work will it? as the module will actually be called ```RulesTest.FakeRule```. It's one of  the reasons I wasn't keen on the auto prefixing. I'd still prefer it if dogma loaded rules based on full module names defined in the sets.

lpil: Why wouldn't that work, sorry?

I see your point, but I don't like having to use fully qualified module names in the config file. Perhaps we should check to see if it is defined under `Dogma.Rule`, first, and then treat it as a full name if not.

meadsteve: well unless I'm mistaken if I move the ```FakeRule``` in to ```Dogma.RulesTest``` the module it will then be called: ```Dogma.RulesTest.FakeRule```.

If I then specify this as ```FakeRule``` dogma tries to load ```Dogma.Rules.FakeRule``` which doesn't exist.

If I specify this as ```RulesTest.FakeRule``` dogma tries to load ```Dogma.Rules.RulesTest.FakeRule``` which also doesn't exist.

It's definitely possible I've got confused however.

lpil: I'm suggesting we call the module `Dogma.Rules.RulesTest.FakeRule` :)

test/dogma/rules_test.exs
```diff
@@ -0,0 +1,67 @@
...
+    single_script = %Script{}
+    expected_script_after_run = %Script{errors: [:always_fake_error]}
+
+    result = Rules.test([single_script], @formater_for_testing)
+
+    assert result == [expected_script_after_run]
+  end
+
+  test "Formatter.script is called by Rules.test" do
+    Application.put_env(:dogma, :rule_set, Dogma.RuleSets.FakeRuleSet)
```
lpil: Is this safe to do? I don't really like changing global state like this.

Is this the right way to be testing this?

lpil:
I'm not sure we're testing the right things here. I'm also not sure this module is doing what we want it to, it seems to have several responsibilities.

meadsteve:
@lpil somehow I forgot to reply to this. I agree this module does seem to have several responsibilities. I'll have a think on what it makes sense to test. At the very least it seems we should be testing that it calls out to all the rules provided by the configured set.

lpil:
Lets leave this for now, I'm going to remove this module in future.

PR #147, opened 2015-10-27
Make formatters feed off of events

### LINKED ISSUE #143: Move away from passing in "formatter" modules to callbacks, genevents, or similar.
We should pass in callback functions rather than a formatter module in order to react to the results of tests.

As suggested in #129

5thWall:
Closes #143


lib/dogma.ex
```diff
@@ -6,20 +6,29 @@ defmodule Dogma do
...
   """
 
-  alias Dogma.Formatter
   alias Dogma.Rules
   alias Dogma.ScriptSources
 
   @version Mix.Project.config[:version]
 
-  def run(dir, config, formatter) do
+  def run(dir, config, events) do
```
lpil: Can we call this `gen_event` or `event_listener` or something like that please? :)

lib/dogma/formatter.ex
```diff
@@ -1,56 +1,11 @@
 defmodule Dogma.Formatter do
```
lpil: Can we call these reporters or something now? They don't just format data any more.

lib/dogma/rules.ex
```diff
@@ -12,16 +11,16 @@ defmodule Dogma.Rules do
   @doc """
   Runs the rules in the current rule set on the given scripts.
   """
-  def test(scripts, rules, formatter) do
+  def test(scripts, rules, events) do
```
lpil: And here.

lib/mix/tasks/dogma.ex
```diff
@@ -11,8 +11,11 @@ defmodule Mix.Tasks.Dogma do
     {dir, formatter} = argv |> parse_args
     config = Config.build
 
+    {:ok, events} = GenEvent.start_link([])
```
lpil: Does this belong in a supervision tree?

lib/mix/tasks/dogma.ex
```diff
@@ -11,8 +11,11 @@ defmodule Mix.Tasks.Dogma do
...
+    GenEvent.add_handler(events, formatter, [])
+
     dir
-    |> Dogma.run(config, formatter)
+    |> Dogma.run(config, events)
```
lpil: And here.

test/dogma/formatter/flycheck_test.exs
```diff
@@ -5,70 +5,37 @@ defmodule Dogma.Formatter.FlycheckTest do
...
-        assert Flycheck.script(script) == ""
-      end
-    end
-
-    with "some errors" do
-      should "print nothing" do
-        script = %Script{ errors: [%Error{}] }
-        assert Flycheck.script(script) == ""
-      end
+  with "no errors" do
```
lpil: We should test the GenEvent functionality as well as the functions that format the strings.

lib/dogma.ex
5thWall: The formatters (or whatever we call them) are really the listeners here, this is more of a dispatcher.  How about `event_manager` or just `dispatcher`. It's referred to as an event manager in the docs. I'm partial to `dispatcher` but that's less explicit.

lib/mix/tasks/dogma.ex
5thWall: It could be in a supervision tree. If we're going to do that then we should give it a global name, since the PID will change when it restarts.

I'm not sure how to go about re-adding event handlers when it restarts, so it might just be best to let it fail for now and tackle that later. I guess it depends, if this crashes would it be better for the entire app to crash, or for the app to proceed and the only indication of ultimate success or failure be the exit code?

lib/dogma.ex
lpil: Both sound good. Your call.

lib/mix/tasks/dogma.ex
lpil: I feel Dogma should probably crash if the reporters fail.

Again, I've not really got into OTP much, so let me ask this. Due to the use of `start_link/2` if the event handler process fails the parent process will also fail, as they are connected together, correct?
If that's the case, this implementation means they both crash, which I think is what we want.

5thWall: Yes, as it stands right now if an event handler fails then the entire app crashes. This could be for something stupid, like not having a default `handle_event/2` function, so we should watch out for that with event handlers.

alxndr: Would it make any sense for the linting to still be successful if something crashes?

5thWall: If one of the reporters crashes, but all of the rules pass then it could makes sense in a CI environment. If none of the rules fail then the build should still be green. But that's a pretty narrow use case; I think for the most part people are going to want to see reports.

lpil: I think if the testing mechanism fails in some fashion on CI then the build should fail.

alxndr: I also think the CI build should fail in that case.

docs/reporters.md
```diff
@@ -1,21 +1,20 @@
-# Dogma Formatters
+# Dogma Reporterss
```
lpil: Typo here :)

lib/dogma/reporter/json.ex
```diff
@@ -34,17 +34,14 @@ defmodule Dogma.Formatter.JSON do
...
   """
 
-  @behaviour Dogma.Formatter
+  use GenEvent
 
-  @doc """
-  Runs at the start of the test suite, printing nothing.
-  """
-  def start(_), do: ""
+  def handle_event({:finish, scripts}, _) do
```
alxndr: Looks like the other modules have `:finished`?

test/dogma/reporter/simple_test.exs
```diff
@@ -0,0 +1,152 @@
...
+    %Script{
+      path: "bar.ex",
+      errors: [
+        %Error{ rule: Foo.Confusing,  line: 2,  message: "Wtf?" },
+        %Error{ rule: Foo.UglyAsHell, line: 63, message: "Not ok." }
+      ]
+    },
+  ]
+
+  with ".handle_event" do
```
5thWall: So, these tests are feeling pretty redundant. What do you think of moving `start/1`, `script/1`, `finish/1` to private, and then just using the event handlers to test the full thing?

lpil: Sounds like a very good idea to me. These functions are internal implementation details.

lpil:
Looking good so far, thanks Andy

lpil:
Hi Andy

Sorry for taking so long to get back to you. I've been in the process of moving house, and it's sucked up all my focus.

Thank you so much for this. It's fantastic!

Cheers,
Louis

5thWall:
I can take a look at it tonight. I've had a lot taking up my focus as well with a new job and moving to a different city, so I probably didn't put my best effort in on the last few commits.

lpil:
That's quite understandable, I'm in the same boat. I might take a look if I get some free time (and internet!).

5thWall:
Had some Dogma violations, but they weren't displayed on the screen because the program would exit before all the reporters finished. I changed all the events to notify synchronously instead of async and that cleared it up.

lpil:
Great! Thank you. :)

PR #153, opened 2015-11-04
Add better string removal

rrrene:
Hi there, long time no see. I was working on a prototype for an Elixir parser/toolbox/linter and noticed [this comment of yours](https://github.com/lpil/dogma/pull/149) about `Dogma.ScriptStrings` still falling short. :sob: 

This PR adds two more edge cases to the test suite where `ScriptStrings` was failing and adds the code I developed for my own project to resolve these cases.  The public API of `ScriptStrings` was *not* altered (I adapted my code) and you should be able to finish your Comment rule soon. :grin:

### DIFF
--- a/lib/dogma/util/script_strings.ex
+++ b/lib/dogma/util/script_strings.ex
@@ -7,7 +7,7 @@ defmodule Dogma.Util.ScriptStrings do
   for Elixir code and reporting errors when there should be none.
   """
 
-  @sigil_delimiters [{"|", "|"}, {"\"", "\""}, {"'", "'"}, {"(", ")"},
+  @sigil_delimiters [{"|", "|"}, {~S("), ~S(")}, {"'", "'"}, {"(", ")"},
                       {"[", "]"}, {"{", "}"}, {"<", ">"}]
   @all_string_sigils @sigil_delimiters
                       |> Enum.flat_map(fn({b, e}) ->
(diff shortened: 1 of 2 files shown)


lpil:
This is fantastic, thank you! :)

lib/dogma/util/script_strings.ex
```diff
@@ -7,98 +7,106 @@ defmodule Dogma.Util.ScriptStrings do
   for Elixir code and reporting errors when there should be none.
   """
 
+  @sigil_delimiters [{"|", "|"}, {~S("), ~S(")}, {"'", "'"}, {"(", ")"},
+                      {"[", "]"}, {"{", "}"}, {"<", ">"}]
+  @all_string_sigils @sigil_delimiters
+                      |> Enum.flat_map(fn({b, e}) ->
+                          [{"~s#{b}", e}, {"~S#{b}", e}]
```
lpil: What about other, non `sS` sigils?

rrrene: I wasn't sure which ones to include. Basically, there seem to be two alternatives: 

1. Include only s/S, so the module does what it says it does ("Takes a source string and return it with all of the string literals stripped of their contents.") or
2. include all the standard sigils (c,r,s,w).

So you basically have to decide whether to be correct or practical :wink: But that is for you to decide, I just implemented what the docstring said.

lpil: I think it would be more useful to remove all sigils, including custom ones. We can rename this module, change docs, etc. If we don't, we still have the same problems with not being able to tell if "comments" are inside sigils.

Anyone got a good name?

rrrene: The way I do it in my project is that I have two modules `Code.Strings` and `Code.Sigils` which handle these things independently. This is easy to comprehend and test and not much of on overhead, since you only seem to use the functionality in a single place.

Anyhow. This PR stands as it is, since it provides what it says in the title and improves the functionality of a given module. I would rather not change course mid-PR. :+1:

lpil: Do you think there is value in merging this in then? Sounds like we'll just undo it in future.

rrrene: You're the maintainer, it is your choice (and responsibility) to decide what goes into the codebase and what doesn't. Me and the other contributors, like @5thWall and @alxndr, are simply trying to add something of value.

That said, `Dogma.Util.ScriptStrings` is broken at the moment. It does simply not do what it promises to do (again "take a source string and return it with all of the string literals stripped of their contents."). This PR fixes it. If you don't want it, because you realized you did want that specific functionality in the first place, that is completely fine.

But "we might change this code in the future to achieve something not described today" is not a valid reason to reject a PR that fixes a current problem, IMHO.

lpil: I agree, thanks Rene.

Issue #33, opened 2015-06-30
Allow a path to a directory or file to be tested to be passed to the mix task as an argument


lpil:
i.e. build upon #32 

With documentation and tests!

lpil:
Closed in #154

PR #154, opened 2015-11-07
Allow linting of single files

tolbrino:
Example: `mix dogma config/config.ex`

Works for `.ex` and `.exs` files

Ref #33

### DIFF
--- a/lib/dogma/script_sources.ex
+++ b/lib/dogma/script_sources.ex
@@ -19,20 +19,15 @@ defmodule Dogma.ScriptSources do
   """
   def find(path, exclude_patterns \\ []) when is_list exclude_patterns do
     path = to_string path # Convert nil to ""
-    case File.regular? path do
-      true ->
-        case String.ends_with? path, [".ex", ".exs"] do
-          true ->
-            [path]
-          false ->
-            []
-        end
+    case resolve_file path do
       false ->
         path
         |> Path.join( "**/*.{ex,exs}" )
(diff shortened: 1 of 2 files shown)


lib/dogma/script_sources.ex
```diff
@@ -17,13 +17,23 @@ defmodule Dogma.ScriptSources do
...
-  def find(dir_path, exclude_patterns \\ []) when is_list exclude_patterns do
-    dir_path
-    |> to_string # Convert nil to ""
-    |> Path.join( "**/*.{ex,exs}" )
-    |> Path.wildcard
-    |> Enum.reject( &String.starts_with?(&1, @ignored_dirs) )
-    |> Enum.reject( &match_any?(&1, exclude_patterns) )
+  def find(path, exclude_patterns \\ []) when is_list exclude_patterns do
+    path = to_string path # Convert nil to ""
+    case File.regular? path do
```
lpil: I don't like all this branching here. Can we extract this into a clearly named function so future readers can understand the intent?

lpil:
A good change, thanks. One note above.

lib/dogma/script_sources.ex
tolbrino: I didn't like my initial approach either. How about the revision?

tolbrino:
@lpil Anything missing here or can this be merged?

lpil:
Thanks!

PR #156, opened 2015-11-09
Fix broken Hex docs link in README

mgartner:
This fixes the broken Hex docs link.


lpil:
Huh. Did the url change?

Thanks Marcus!

PR #157, opened 2015-11-09
New rule: FunctionParentheses

mgartner:
I added a new rule which ensures parentheses are used when a function declaration has arguments, and not used when it doesn't have arguments. I figured that this is a good rule to have as it is listed in this Elixir Style Guide: https://github.com/niftyn8/elixir_style_guide#syntax. Let me know if I can improve my implementation in any way - I'm new to Elixir.

For example:

```elixir
# These are good:

def foo do
end

def bar(x) do
end

# These are bad:

def foo() do
end

def bar x do
end
```

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -1,7 +1,7 @@
 # Dogma Rules
 
 These are the rules included in Dogma by default. Currently there are
-27 of them.
+28 of them.
 
 ## Contents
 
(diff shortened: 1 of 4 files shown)


mgartner:
It looks like the build fails for Elixir 1.0.5. I'll have to look into why that is.

lpil:
Thanks Marcus

There are some difference between the AST in version v1.0 and v1.1 which could be the problem. Would be good to hear what you find. :)

mgartner:
Looks like the AST in 1.1 includes character positioning in addition to line location, changing the structure from `[{:identifier, 1, :foo}]` to `[{:identifier, {1, 4, 7}, :foo}]`.

lpil:
Ah, yes. We handle it like so: https://github.com/lpil/dogma/blob/master/lib/dogma/rule/exception_name.ex#L64

mgartner:
Thanks for the tip!

Not sure how I made coverage go down. Any ideas?

lpil:
It's happened on all pull requests lately. I think that the way that coverage is calculated has changed somehow.

lpil:
Thanks for this! A great addition :)

PR #162, opened 2015-11-11
New rule: MultipleBlankLines

mgartner:
This rule prevents multiple consecutive blank lines. I've found this type of rule useful in linters for other languages. I'm curious to see whether others agree with this rule or not in Elixir-land.

### DIFF
--- a/test/dogma/config_test.exs
+++ b/test/dogma/config_test.exs
@@ -61,8 +61,9 @@ defmodule Dogma.ConfigTest do
     end
 
     should "default to []" do
-      assert nil == Application.get_env( :dogma, :rule_set )
-      assert [] == Config.build.exclude
+      TemporaryEnv.delete( :dogma, :exclude ) do
+        assert [] == Config.build.exclude
+      end
     end
   end
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,1 +1,5 @@
 use Mix.Config
+
+config :dogma,
+  rule_set: Dogma.RuleSet.All,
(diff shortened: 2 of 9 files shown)


lpil:
Can we have this rule accept a configurable number of max lines? For this project I would like to limit it to (but allow) 2 newlines in a row.

lib/dogma/documentation/rules_list.ex
```diff
@@ -34,7 +34,6 @@ defmodule Dogma.Documentation.RuleList do
     """
   end
 
-
```
lpil: Like here, I would like to keep these newlines. :)

lpil:
A good rule, thanks for the contribution. :)

mgartner:
Good idea I'll work on adding the ability to customize this rule.

mgartner:
@lpil added configuration for this rule, and cleaned up the code a bit. I left a handful of blank line removals that I suspect you don't want to keep, but lmk if you do and I can add them back.

test/dogma/config_test.exs
```diff
@@ -61,8 +61,9 @@ defmodule Dogma.ConfigTest do
     end
 
     should "default to []" do
-      assert nil == Application.get_env( :dogma, :rule_set )
-      assert [] == Config.build.exclude
+      TemporaryEnv.delete( :dogma, :exclude ) do
```
lpil: Yay :)

lpil:
Looks great, thank you! :)

PR #165, opened 2015-11-13
New rule: InfixOperatorSpaces

mgartner:
This is an initial idea of how to enforce that spaces are put around infix operators. It "works" by finding the lines and locations of each infix operator, then checking the characters before and after those operators to ensure that they are spaces. There's a couple problems with this approach:

1. It relies on the position information of the infix operators given in the `:tokens` list. This information is not available in Elixir 1.0.5  as far as I know.
2. For some reason, unary operators (i.e. the `-` in `a = -1`) are given the token type of `:dual_op` in the list of tokens, when it seems like they should be listed as `:unary_op`. This is currently causing the tests to fail because I can't think of an easy way for the rule to recognize that there shouldn't be a space after unary operators like this.

Any ideas on how to resolve the two issues above? Maybe there's a smarter approach entirely?

### DIFF
--- a/test/dogma/rule/infix_operator_spaces_test.exs
+++ b/test/dogma/rule/infix_operator_spaces_test.exs
@@ -1,0 +1,83 @@
+defmodule Dogma.Rule.InfixOperatorSpacesTest do
+  use ShouldI
+
+  alias Dogma.Rule.InfixOperatorSpaces
+  alias Dogma.Script
+  alias Dogma.Error
+
+  defp lint(script) do
+    script |> Script.parse!( "foo.ex" ) |> InfixOperatorSpaces.test
+  end
+
+  should "not error with spaces around infix operators" do
+    errors = """
+    1 < 2 && 2 >= 3 || 4 > 7 || 3 <= 9
+    2 - 1 * 3 / 4 + 5
+    a = 1
+    b ==
+      5
+    %{:c => 1}
+    """ |> lint
+    assert [] == errors
+  end
+
+  should "not error with unary operators" do
+    errors = """
+    -1
+    b = -3
+    !c
+    """ |> lint
+    assert [] == errors
+  end
+
+  should "error without spaces before infix operators" do
+    errors = """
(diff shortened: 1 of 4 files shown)


mgartner:
Maybe a better approach is to use a regex on each item in the `processed_lines` list?

lpil:
Hello Marcus, sorry about the slow reply. I've had a manic couple of months.

I'd like to support 1.0.5, but I'm not sure what the right approach to this problem is...

How would you use regex here?

mgartner:
Maybe something like running testing these two regexes on every line and erring if a match is found.

```elixir
~r/\w(\+|\*|=|<|&|\|)/
~r/(\+|\*|=|<|&|\|)\w/
```

I'll try to figure out if this will work and throw together a PR.

test/dogma/rule/infix_operator_spaces_test.exs
```diff
@@ -0,0 +1,83 @@
...
+    -1
+    b = -3
+    !c
+    """ |> lint
+    assert [] == errors
+  end
+
+  should "error without spaces before infix operators" do
+    errors = """
+    1+ 2
```
alxndr: how about `2- 1`?

mgartner:
@lpil I'd love to take another stab at this, but I'm not sure it's really possible in Elixir v1.0.5 or less because it lacks position information for each token. What do you think of adding a rule that only is enabled if the Elixir version is v1.1 or greater?

lpil:
An interesting idea. How do you envision version specific rules working?

mgartner:
I haven't thought much about it, but I could either report no errors if `System.version/0` returns less that `"1.1.x"` or create separate rule sets and select a set based on `System.version/0`.

mgartner:
It looks like credo is able to apply this rule on Elixir 1.0.5, so I'll first look into how that works and see if it can apply here.

mgartner:
I'm going to take a new approach on this. I'll resubmit a new PR when ready.

PR #166, opened 2015-11-16
Remove weird exception.

jonnyarnold:
The rest of the settings in the default rule set are based on personal preference. However, Dogma sets an override to *its own rules* for the maximum number of blank lines, which doesn't make sense to me - why is this opinion not strong enough to be considered part of the default rule set?

This PR ensures the opinions of the Dogma writers are entirely in the default rule set.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,5 +1,4 @@
 use Mix.Config
 
 config :dogma,
-  rule_set: Dogma.RuleSet.All,
-  override: %{ MultipleBlankLines => [ max_lines: 2 ] }
+  rule_set: Dogma.RuleSet.All
--- a/lib/dogma/rule_set/all.ex
+++ b/lib/dogma/rule_set/all.ex
@@ -26,7 +26,7 @@ defmodule Dogma.RuleSet.All do
       ModuleAttributeName     => [],
       ModuleDoc               => [],
       ModuleName              => [],
-      MultipleBlankLines      => [max_lines: 1],
+      MultipleBlankLines      => [max_lines: 2],
       NegatedAssert           => [],
       NegatedIfUnless         => [],
       PipelineStart           => [],


lpil:
I'm happy to merge this in, but I presumed that most people would want this to have 1 max empty line.

Anyone have any preference? @rrrene @jacknoble @meadsteve @alxndr @tolbrino @5thWall

rrrene:
I regularly use 2-3 blank lines in larger modules (tests for example) to separate parts of the code ... But this is just an opinion and I'm sure 1 blank line as default is fine. People will customize it anyway :wink:

meadsteve:
What's the longest in the elixir lang source? Makes sense to consider that the "standard".

asaaki:
> People will customize it anyway

I'd like not to be "forced" to customize it, if the default is a reasonable value.
From my experience in many language linters more than a single empty line is considered an offense/error/code smell.

What does a second or third blank line serve?

jacknoble:
I agree that most people will want 1 line and that therefore that should be the default.

tolbrino:
+1 for 1 line

meadsteve:
The other thing I'd say is that having an exception in dogma itself does seem weird. I'm looking at the defaults as the ```dogma standard```. To me dogma itself should be the example of the standard.

jonnyarnold:
@meadsteve Exactly what I was suggesting - a linter that makes exceptions to its own rules is a little weird.

lpil:
I'm not willing to stop using 2 lines to break things up into logical paragraphs, especially in tests.

>  a linter that makes exceptions to its own rules is a little weird.

Well, the idea is that this tool is supposed to be customisable.

meadsteve:
@lpil I'd argue if you aren't willing to remove 2 line breaks then this should be the default. I don't think there's anything wrong with this. For a lot of the rules the decisions are fairly arbitrary and we may as well pick what is already the case in this repo (unless there's a very good reason not to).

Unless a style guide actively advocates never having more than 1 line break I don't see any problems.

So I'm in favour of merging this PR.

lpil:
OK, lets do it then. Thanks all.

alxndr:
:+1:

PR #167, opened 2015-11-18
Dogma should not be a production dependency

johnhamelink:
Hey there,

I noticed that the readme recommends you install Dogma only specifying that it should be run in development  - to me it seems that Dogma should also be run in the test environment also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Add Dogma to your Mix dependencies
 # mix.exs
 def deps do
   [
-    {:dogma, "~> 0.0", only: :dev},
+    {:dogma, "~> 0.0", only: [:dev, :test]},
   ]
 end
 ```


lpil:
Why is this? Are you using Dogma in your tests somehow?

Could you change the commit message to be accurate please?

Thanks :)

johnhamelink:
@lpil Thanks for getting back to me so quick!

> Are you using Dogma in your tests somehow?
I could be using Dogma in CI where I would be running the code in a test environment - this was the primary use case I had in mind.

> Could you change the commit message to be accurate please?
Done

lpil:
I feel this is the responsibility of the person running Dogma The minimum required to run Dogma is for it to be in the `:dev` env.

johnhamelink:
@lpil I agree, but disagree that the minimum is the most helpful for new users in this case. I'll close this PR.

Thanks for Dogma, keep up the great work, I hope I'll be able to contribute to it in a more meaningful way sometime in the near future! :rocket:

lpil:
Thank you John. If you think this is something worth documenting, maybe we should add a new docs page that contained information on running Dogma on CI :)

Cheers!

Issue #171, opened 2015-12-13
What if rules were structs?


lpil:
The way in which we currently pass around the names of Rule modules seems kind of awkward to me, I wonder if we can do better. What if rules were structs?

```elixir
defmodule Dogma.Rule.LineLength do
  defstruct enabled: true,
            max_length: 80
end

defimpl Rule, for: Dogma.Rule.LineLength do
  @alias Dogma.Rule.LineLength
  @alias Dogma.Script

  def test(rule, %Script{} = script) do
    # perform the test
  end
end
```

The we just iterate over the list of rule structs and call the `Rule.test/2` function on each one.

We could write a `defrule` macro to cut down on the boilerplate.

```elixir
defrule LineLength, [max_length: 80] do
  def test(rule, %Script{} = script) do
    # perform test
  end
end
```

This would also solve the problem of having to either have very verbose names in the mix config, or concatenating `Dogma.Rule` on the start of every name atom.

meadsteve:
I like this. The struct gives a really clear structure for what config options are available. It even makes it technically possible for IDEs to autocomplete on this. 

I'm not sure if this does solve the ```very verbose names in the mix config, or concatenating Dogma.Rule on the start of every name atom.``` problem however. The struct will still have the verbose name right?

meadsteve:
Also just realised it means you can do things like
```elixir
rule_structs
  |> Enum.filter(fn rule -> rule.enabled end)
```
to get all the enabled rules. Seems a sensible way of doing this.

lpil:
Well, you can alias the patent module, that helps a little, and is much clearer than the previous solution.

Would be good to get a review in the PR once I've made some progress. :)

meadsteve:
@lpil no problem. Let me know when you'd like me to take a look (random unavailability due to travelling around over xmas caveat applies though).

lpil:
Merry everybody :santa:

PR #172, opened 2015-12-13
Dispatch to rules using a protocol.

lpil:
See #171 

Breaking changes ahoy.

- [x] `Dogma.Rule` protocol
- [x] `defrule` macro
- [x] `Rule.Case` test macro
- [x] All rules using `defrule`
- [x] All rules tested using `Rule.Case`
- [x] Rule sets updated to use structs
- [x] Dogma using protocol to run rules
- [x] Config generation handles new format
- [x] Config supports old format
- [x] Deprecation message for old format
- [x] Config documentation updated
- [x] No pending tests
- [ ] Changelog


lib/dogma/rule/comment_format.ex
```diff
@@ -1,4 +1,6 @@
-defmodule Dogma.Rule.CommentFormat do
+use Dogma.RuleBuilder
+
+defrule Dogma.Rule.CommentFormat, [] do
```
meadsteve: small comment but if there are no default config values maybe 
```elixir
defrule Dogma.Rule.CommentFormat do
```
would be better.

lpil: Yeah, I should do that.

lpil:
@meadsteve 

Care to review?

Also, do you know how to go from a module atom name to the struct for that module?

meadsteve:
@lpil I think the kernel function ```struct(struct, kv \\ []) ```might be what you're looking for. I'll take a look at this pr today when I get a chance.

lpil:
Ah, thanks! I didn't think  to look in kernel. :)

lib/dogma/config.ex
```diff
@@ -25,17 +25,46 @@ defmodule Dogma.Config do
...
-    Dict.delete( rules, rule )
+
+  defp insert_rule(rule, acc) when is_map(rule) do
+    Dict.put(acc, rule.__struct__, rule)
   end
-  defp reconfigure_rule({rule, config}, rules) do
-    Dict.put( rules, rule, config )
+
+  defp insert_rule({rule_name, config}, acc)
+  when is_atom(rule_name) and is_list(config) do
```
meadsteve: :+1: like this approach to updating the config style. Neat way of emitting deprecations and the whole method can just be removed at the next major version bump.

One thing I'd think about though is whether it'd be better to do this as a logger.warn rather than just an IO.put. As it is a warning not just a message.

lib/dogma/rule_builder.ex
```diff
@@ -0,0 +1,62 @@
...
+      require Dogma.RuleBuilder
+      import Dogma.RuleBuilder, only: [defrule: 3, defrule: 2]
+    end
+  end
+
+  defmacro defrule(name, [do: module_ast]) do
+    quote do
+      defrule unquote(name), [], do: unquote(module_ast)
+    end
+  end
```
meadsteve: Is the quote here needed? I'd have thought 
```elixir
  defmacro defrule(name, [do: module_ast]) do
    defrule(name, [], module_ast)
  end
```
assuming this is just used to provide default opts as []

lib/dogma/config.ex
lpil: Currently the logger isn't running. I could add it but then I need to backport `ExUnit.CaptureLog`, as it isn't present in 1.0 :(

lib/dogma/rule_builder.ex
lpil: Seems to be required.

```elixir
  defmacro defrule(name, [do: module_ast]) do
    defrule name, [], do: module_ast
  end
```

```
== Compilation error on file lib/dogma/rule_builder.ex ==
** (CompileError) lib/dogma/rule_builder.ex:37: undefined function defrule/3
    (stdlib) lists.erl:1337: :lists.foreach/2
    (stdlib) erl_eval.erl:669: :erl_eval.do_apply/6
```

lib/dogma/config.ex
meadsteve: ah I see. That's a shame. In any case what you've done should still be useful.

lib/dogma/rule_builder.ex
```diff
@@ -0,0 +1,62 @@
...
+  defmacro defrule(name, opts, [do: module_ast]) when is_list(opts) do
+    opts = [{:enabled, true} | opts]
+    quote do
+      defmodule unquote(name) do
+        alias Dogma.Error
+        alias Dogma.Script
+
+        defstruct unquote(opts)
+        unquote(module_ast)
+
```
meadsteve: what happens if test/2 isn't defined? Does it error on compile? (please ignore all of the following if it does error). One thought is to add a rule @behaviour so this starts something like:
```elixir
quote do
   defmodule unquote(name) do
      alias Dogma.Error
      alias Dogma.Script
 
       @behaviour Dogma.RuleBehaviour
```

lib/dogma/rule_builder.ex
meadsteve: Maybe
```elixir
defmacro defrule(name, [do: module_ast]) do
   rule_ast name, [], do: module_ast
end

defmacro defrule(name, opts, [do: module_ast]) do
   rule_ast name, opts, do: module_ast
end

defp rule_ast(name, opts, [do: module_ast]) when is_list(opts) do
#...
```
I think you should be able to call a function from a macro

meadsteve:
A few comments from me but generally this approach looks really good. I like that the script module got smaller and more focussed.

lib/dogma/rule_builder.ex
lpil: It compiles and crashes at runtime. I'll add the behaviour so it compiles with a warning.

lib/dogma/rule_builder.ex
lpil: I'm not sure this is any clearer than the quote. I'll stick with this :)

lpil:
Thanks Steve!

lib/dogma/rule_builder.ex
meadsteve: Yeah it just seemed odd quoting a macro. I wasn't quite sure what was happening. But it works and has tests around it so I wouldn't be too worried.

PR #174, opened 2015-12-17
Fix dead link to formatters in the Readme

pjambet:
Seems like #147 broke this link by renaming the file from formatters to
reporters.
I kept the link text as 'Output formats' as it seems like the CLI option
is still called format even though the module is now called reporters.


lpil:
Thank you.

Issue #173, opened 2015-12-15
Improve HardTabs rule


lpil:
Detect tabs that come after spaces but before code.

```elixir
"""
  # spaces here are ok
     \tIO.puts("Hi") # spaces then a tab, not ok!
"""
```

lpil:
Closed by @msoedov  :)

PR #179, opened 2016-01-24
#173 Add rule for mixed tabs and spaces

msoedov:
#173 Modified regular expression for HardTabs rule to detect code indented with a mixture of tabs and spaces

### DIFF
--- a/lib/dogma/rule/hard_tabs.ex
+++ b/lib/dogma/rule/hard_tabs.ex
@@ -11,7 +11,7 @@ defmodule Dogma.Rule.HardTabs do
 
   @behaviour Dogma.Rule
 
-  @indenting_tab_pattern ~r/^\t+.+/
+  @indenting_tab_pattern ~r/^[\t\s]*\t+/
 
   alias Dogma.Error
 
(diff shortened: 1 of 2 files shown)


lib/dogma/rule/hard_tabs.ex
```diff
@@ -11,7 +11,7 @@ defmodule Dogma.Rule.HardTabs do
 
   @behaviour Dogma.Rule
 
-  @indenting_tab_pattern ~r/^\t+.+/
+  @indenting_tab_pattern ~r/^(\s*)[\t\s]*\t+/
```
lpil: What is the first capture group for? Couldn't we remove that?

lpil:
Thanks for the PR! One comment above.

lib/dogma/rule/hard_tabs.ex
msoedov: good call

lpil: Sorry, might be me being slow, but what do we need the first `\s*` for?

lpil:
Sorry, might be me being slow, but what do we need the first \s* for?

msoedov:
@lpil  Good caught

lpil:
Great! Thank you Alex :)

Issue #146, opened 2015-10-24
New rule SpaceAfterComma


lpil:
Enforce spaces after commas. Or no spaces. Or `n` spaces. Or `x..y` spaces.

```elixir
[1,2,3,4]
[1, 2, 3, 4]
[1,  2,  3,  4]
```

mgartner:
I believe this can be closed @lpil.

lpil:
Thank you.

PR #184, opened 2016-02-21
SpaceAfterComma rule

msoedov:
#146 That's a really straightforward implementation. However there are 39 errors! for this rule in the project. Some of those are really easy to fix but there are couple places where spaces after comma used for some kind of visual indent like

```elixir
  defp elixirc_paths(:test), do: ["lib", "test/support"]
  defp elixirc_paths(_),     do: ["lib"]
```

What should we do with such corner cases?

### DIFF
--- a/lib/dogma/rule/space_after_comma.ex
+++ b/lib/dogma/rule/space_after_comma.ex
@@ -11,6 +11,7 @@ defrule Dogma.Rule.SpaceAfterComma, [spaces: 1] do
 
   def test(rule, script) do
     script.lines
+    |> Enum.map(&drop_strings/1)
     |> Enum.filter(fn line -> check_line(rule, line) end)
     |> Enum.map(fn line ->  error(rule, line) end)
   end
@@ -20,14 +21,18 @@ defrule Dogma.Rule.SpaceAfterComma, [spaces: 1] do
     |> Regex.scan(line)
     |> Enum.any?(fn [_, group] ->
       String.length(group) != rule.spaces
-      end)
+    end)
+  end
+
+  defp drop_strings({pos, line}) do
+    {pos, Regex.replace(~r/\"(.*?)\"/, line, "\"\"")}
   end
 
   defp error(rule, {pos, _}) do
     %Error{
       rule:     __MODULE__,
       message:  "Should be #{rule.spaces} spaces after comma",
-      line: Dogma.Script.line(pos),
+      line:     Dogma.Script.line(pos),
     }
   end
 end
--- a/lib/dogma/rule/space_after_comma.ex
+++ b/lib/dogma/rule/space_after_comma.ex
@@ -1,0 +1,22 @@
+use Dogma.RuleBuilder
+
+defrule Dogma.Rule.SpaceAfterComma, [spaces: 1] do
+  @moduledoc """
+  A rule that enforces to use `spaces` after comma.
+
+  This rule can be configured with the `spaces` option, which allows you to
+  specify your own number of spaces after comma.
+  """
+
+  def test(rule, script) do
+    []
+  end
+
+  defp error(rule, {pos, line}) do
+    %Error{
+      rule:     __MODULE__,
+      message:  "Should be #{rule.spaces} spaces after comma",
+      line: Dogma.Script.line(pos),
+    }
+  end
+end
--- a/lib/dogma/rule/space_after_comma.ex
+++ b/lib/dogma/rule/space_after_comma.ex
@@ -2,7 +2,7 @@ use Dogma.RuleBuilder
 
 defrule Dogma.Rule.SpaceAfterComma, [spaces: 1] do
   @moduledoc """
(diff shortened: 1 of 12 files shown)


lpil:
I think that we should enforce >= 1 spaces by default, with configurable modes for = 1 spaces, and =0 spaces.

lib/dogma/rule/space_after_comma.ex
```diff
@@ -0,0 +1,33 @@
...
+    |> Enum.map(fn line ->  error(rule, line) end)
+  end
+
+  defp check_line(rule, {_, line}) do
+    @spaces_pattern
+    |> Regex.scan(line)
+    |> Enum.any?(fn [_, group] ->
+      String.length(group) != rule.spaces
+      end)
+  end
```
lpil: Can we indent this like this please?

```elixir
defp check_line(rule, {_, line}) do
  @spaces_pattern
  |> Regex.scan(line)
  |> Enum.any?(fn [_, group] ->
    String.length(group) != rule.spaces
  end)
end
```

lib/dogma/rule/space_after_comma.ex
```diff
@@ -0,0 +1,33 @@
...
+
+  defp error(rule, {pos, _}) do
+    %Error{
+      rule:     __MODULE__,
+      message:  "Should be #{rule.spaces} spaces after comma",
+      line: Dogma.Script.line(pos),
```
lpil: Indent this value to match the others please.

msoedov:
@lpil that's a good point, I'll updated it today

lib/dogma/rule/space_after_comma.ex
```diff
@@ -0,0 +1,33 @@
...
+  @moduledoc """
+  A rule that enforces to use `spaces` after comma.
+
+  This rule can be configured with the `spaces` option, which allows you to
+  specify your own number of spaces after comma.
+  """
+  @spaces_pattern ~r/\,(\s*).+/
+
+  def test(rule, script) do
+    script.lines
```
lpil: This will throw errors for commas inside strings.

test/dogma/rule/space_after_comma_test.exs
```diff
@@ -0,0 +1,62 @@
+defmodule Dogma.Rule.SpaceAfterCommaTest do
+  use RuleCase, for: SpaceAfterComma
+
+  defp lint(script, rule) do
+    script |> Script.parse!( "" ) |> fn s -> Rule.test(rule, s) end.()
```
lpil: Can we use the testing style used in the other rule tests please?

test/dogma/rule/space_after_comma_test.exs
```diff
@@ -0,0 +1,62 @@
...
+    assert [] == errors
+  end
+
+  defp error_on_line(line, spaces) do
+    %Error{
+      line: Dogma.Script.line(line),
+      message: "Should be #{spaces} spaces after comma",
+      rule: SpaceAfterComma
+    }
+  end
```
lpil: Add a test for commas inside strings.

lpil:
This is looking good, thanks! :)

lib/dogma/rule/space_after_comma.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  This rule can be configured with the `spaces` option, which allows you to
+  specify your own number of spaces after comma.
+  """
+  @spaces_pattern ~r/\,(\s*).+/
+  @string_pattern ~r/\"(.*?,.*?)\"/
+
+  def test(rule, script) do
+    script.lines
+    |> Enum.map(&drop_strings/1)
```
msoedov: Dunno if it possible to ignore string in a better way.

msoedov:
Current version also checks module docks and comments which is kinda side effect but I think should be fine.

lib/dogma/rule/space_after_comma.ex
rrrene: I think @lpil wanted you to use the `processed_source` instead of the `source` of the Script. It is set via [sigils](https://github.com/lpil/dogma/blob/master/lib/dogma/util/script_sigils.ex) and [strings](https://github.com/lpil/dogma/blob/master/lib/dogma/util/script_strings.ex) helpers and [should not contain any sigils and strings](https://github.com/lpil/dogma/blob/410d88070a511e872d340e88f879a15385962a3e/lib/dogma/script.ex#L89) :+1:

lpil: That's the one, sorry I wasn't clear. Thanks Rene! :)

msoedov:
Any other thoughts? Is there are any other corner cases?

lpil:
Fantastic work! Thank you so much :)

lpil:
Nice looking code too

msoedov:
> Nice looking code too

I never feel right when I'm using regexp but thank you!

lpil:
Right tool for the job I think!

Issue #183, opened 2016-02-13
Dogma override config throwing "cannot expand struct" error


elliotlarson:
I'm using Dogma `0.1.0`.  I've added the configuration from the documentation to my `config/config.exs` file:

```elixir
alias Dogma.Rule

config :dogma,

  # Select a set of rules as a base
  rule_set: Dogma.RuleSet.All,

  # Pick paths not to lint
  exclude: [
    ~r(\Alib/vendor/),
  ],

  # Override an existing rule configuration
  override: [
    %Rule.LineLength{ max_length: 120 },
    %Rule.HardTabs{ enabled: false },
  ]
```

When I run:

```bash
$ mix dogma
```

... I get the following error:

> ** (Mix.Config.LoadError) could not load config config/config.exs
    ** (CompileError) config/config.exs:45: Dogma.Rule.LineLength.__struct__/0 is undefined, cannot expand struct Dogma.Rule.LineLength
      (elixir) src/elixir_map.erl:58: :elixir_map.translate_struct/4
      (stdlib) lists.erl:1353: :lists.mapfoldl/3
      (stdlib) lists.erl:1354: :lists.mapfoldl/3

lpil:
Hey. Thanks for the report. Is this project open source, so I can take a look?

elliotlarson:
Sure.  It's just a faux blog learning Elixir/Phoenix project.  Here's the url: https://bitbucket.org/onehouse/blog_phoenix/overview

lpil:
Thank you, I'll take a look when I can.

lpil:
Fixed by @neerfri :D

lpil:
Thanks @elliotlarson

aaronjensen:
fwiw, this did not fix it for us:

```
$ mix dogma
** (Mix.Config.LoadError) could not load config config/dogma.exs
    ** (CompileError) config/dogma.exs:13: Dogma.Rule.LineLength.__struct__/0 is undefined, cannot expand struct Dogma.Rule.LineLength
    (elixir) src/elixir_map.erl:58: :elixir_map.translate_struct/4
    (stdlib) lists.erl:1353: :lists.mapfoldl/3
    (stdlib) lists.erl:1354: :lists.mapfoldl/3
```

aaronjensen:
(we are on 0.1.2)

PR #187, opened 2016-02-26
add support for config/dogma.exs file

neerfri:
When running `mix dogma` with a custom configuration for a rule,
`config.exs` is loaded in mix before dogma's rules are expended so mix
failes with: `config/config.exs:45: Dogma.Rule.LineLength.struct/0 is
undefined, cannot expand struct Dogma.Rule.LineLength` (see
lpil/dogma#183)

This commit allows to put the dogma configuration in `config/dogma.exs`
which is loaded in the beginning of the `mix dogma` task, at this time
the rules are already defined


lpil:
Ah! Thank you so much!

PR #190, opened 2016-02-27
fixes rules are running even when `enabled: false` is set in the config

neerfri:
Before this PR, rules are checked even if they are marked as `enabled: false` in the config.

for example:
```elixir
#config/dogma.exs
use Mix.Config
alias Dogma.Rule

config :dogma,
  rule_set: Dogma.RuleSet.All,
  exclude: [ ],
  override: [
    %Rule.ModuleDoc{ enabled: false },
  ]
```
will still show errors about modules without `@moduledoc`


lpil:
Oh dear. We clearly need more tests. :(

Thanks!

PR #192, opened 2016-02-27
Added a config file fallback chain.

smeevil:
It will now look if it can find config/dogma.exs in the project,
if not it will check if it can find ~/.dogma.exs. otherwise use the defaults

In this way a projects dogma config will always override the user config.

As this is file based, I'm not sure how to write a test for it
I did test this manually though using all four options being :

* No config file at all
* A config in config/dogma.exs
* A config in ~/.dogma.exs
* A config in config/dogma.exs and ~/.dogma.exs where the one in config wins over the home dir one

Hope you'll find this useful :)


smeevil:
Might you have an idea on how to write a code test for this ?

Issue #193, opened 2016-02-29
Accept stdin as input


smeevil:
Hi,

For external editors it would be very nice if Dogma would accept source code from stdin.
This will allow for a lot of IDE's to use Dogma as an inline lint-ing tool, for example with Spacemacs and Flycheck. Currently you need to feed Dogma an existing file to lint.

I've been working on a small prove of concept to see if I could make this happen and here is the result :

![example](https://dl.dropboxusercontent.com/s/jvnc14ge4pqupua/2016-02-29%20at%2019.38.png)

(note the /tmp/test.ex is a dummy value, not an actual file)

Before I dive deeper into this and make everything nice and tidy, I was wondering what you think about this feature and if it has any chance of being merged.

Let me know,

Gerard

lpil:
I didn't see the point in this, but you've given a very good use case here. I'm game. :)

lpil:
Perhaps it should use a special stdin CLI flag? Or another task?

smeevil:
Cool,

Spend quite some time with it today, and almost have everything nice and tidy.
It is indeed using a new flag. 
Will send you a pull request tonight, or maybe earlier if I can sneak some more time in between work :)

Gerard.

lpil:
Thanks Gerard!

PR #194, opened 2016-03-01
Added an option to let dogma parse source from stdin

smeevil:
This pull request will enable dogma to parse source from stdin.
see discussion : https://github.com/lpil/dogma/issues/193

### DIFF
--- a/lib/dogma.ex
+++ b/lib/dogma.ex
@@ -8,19 +8,42 @@ defmodule Dogma do
 
   alias Dogma.Rules
   alias Dogma.ScriptSources
+  alias Dogma.Script
 
   @version Mix.Project.config[:version]
 
   def run(dir_or_file, config, dispatcher) do
     dir_or_file
-    |> ScriptSources.find(config.exclude)
-    |> ScriptSources.to_scripts
+    |> get_sources(config)
     |> notify_start(dispatcher)
     |> Rules.test(config.rules, dispatcher)
     |> notify_finish(dispatcher)
   end
 
+  def get_sources(dir_or_file, %{read_stdin: true} = _config) do
+    if dir_or_file == nil, do: dir_or_file="stdin"
+    case read_from_stdin do
+      {:ok, source} -> [source |> Script.parse(dir_or_file)]
+      _ -> System.halt(666)
+    end
+  end
+
+  def get_sources(dir_or_file, config) do
+    dir_or_file
+    |> ScriptSources.find(config.exclude)
+    |> ScriptSources.to_scripts
+  end
+
   def version, do: @version
+
+  defp read_from_stdin(source \\ "") do
+    case IO.read(:stdio, :line) do
+      {:error, reason} -> {:error, reason}
+      :eof             -> {:ok, source}
+      data             -> source = source <> data
+        read_from_stdin(source)
+    end
+  end
 
   defp notify_start(scripts, dispatcher) do
     GenEvent.sync_notify(dispatcher, {:start, scripts})
--- a/lib/dogma.ex
+++ b/lib/dogma.ex
@@ -39,9 +39,16 @@ defmodule Dogma do
   defp read_from_stdin(source \\ "") do
     case IO.read(:stdio, :line) do
       {:error, reason} -> {:error, reason}
-      :eof             -> {:ok, source}
+      :eof             -> source |> validate
       data             -> source = source <> data
         read_from_stdin(source)
+    end
+  end
+
+  defp validate(source) do
+    case Code.string_to_quoted(source, line: 1) do
+      {:ok, _}        -> {:ok, source}
+      {:error, error} -> {:error, error}
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/dogma.ex
```diff
@@ -8,20 +8,43 @@ defmodule Dogma do
...
+  def get_sources(dir_or_file, config) do
+    dir_or_file
+    |> ScriptSources.find(config.exclude)
+    |> ScriptSources.to_scripts
+  end
+
   def version, do: @version
 
+  defp read_from_stdin(source \\ "") do
+    case IO.read(:stdio, :line) do
```
lpil: Is there an argument to read entirely to EOF in one go?

lib/dogma.ex
```diff
@@ -8,20 +8,43 @@ defmodule Dogma do
...
-    |> ScriptSources.to_scripts
+    |> get_sources(config)
     |> notify_start(dispatcher)
     |> Rules.test(config.rules, dispatcher)
     |> notify_finish(dispatcher)
   end
 
+  def get_sources(dir_or_file, %{read_stdin: true} = _config) do
+    if dir_or_file == nil, do: dir_or_file="stdin"
+    case read_from_stdin do
```
lpil: Can we either use ()s or pass in the accumulator to this function to make it clear it is a call

lib/dogma.ex
```diff
@@ -8,20 +8,43 @@ defmodule Dogma do
...
-    |> ScriptSources.find(config.exclude)
-    |> ScriptSources.to_scripts
+    |> get_sources(config)
     |> notify_start(dispatcher)
     |> Rules.test(config.rules, dispatcher)
     |> notify_finish(dispatcher)
   end
 
+  def get_sources(dir_or_file, %{read_stdin: true} = _config) do
+    if dir_or_file == nil, do: dir_or_file="stdin"
```
lpil: Don't define vars inside blocks and then use them outside of said blocks.

lib/dogma.ex
```diff
@@ -8,20 +8,43 @@ defmodule Dogma do
...
     dir_or_file
-    |> ScriptSources.find(config.exclude)
-    |> ScriptSources.to_scripts
+    |> get_sources(config)
     |> notify_start(dispatcher)
     |> Rules.test(config.rules, dispatcher)
     |> notify_finish(dispatcher)
   end
 
+  def get_sources(dir_or_file, %{read_stdin: true} = _config) do
```
lpil: If we don't use _config, don't assign it

lib/dogma/config.ex
```diff
@@ -15,10 +16,11 @@ defmodule Dogma.Config do
   @doc """
   Build the config struct for the current project.
   """
-  def build do
+  def build(settings \\ %{read_stdin: false}) do
     %__MODULE__{
       rules:   get_rules,
       exclude: get_exclude,
+      read_stdin:   settings[:read_stdin]
```
lpil: Funny spacing here

lib/mix/tasks/dogma.ex
```diff
@@ -41,7 +41,7 @@ defmodule Mix.Tasks.Dogma do
       Reporters.default_reporter
     )
 
-    {List.first(files), reporter, noerror}
+    {List.first(files), reporter, noerror, read_stdin?}
```
lpil: This tuple is starting to get a big unwieldy I think. Perhaps something we could fix in another pull request.

lpil:
How might we be able to test this?

lib/dogma.ex
smeevil: I was thinking to read the source in one go so its easier to add a check later to detect if the source is valid before we hand it off for parsing ?

lib/dogma.ex
smeevil: passing the accumulator now.

lib/dogma.ex
smeevil: Not sure what you mean here...

lib/dogma.ex
smeevil: removed

lib/dogma/config.ex
smeevil: fixed

lib/mix/tasks/dogma.ex
smeevil: agreed

smeevil:
I have no idea how to test it.... :/
All I can say is , it does work when using it.

obmarg:
Just an idea, but maybe this could be generalised a bit so that dogma supported reading from files, with stdin being a special type of file.  That way you could test most of this change by passing it some files.  Might not fully test the stdin case - but it should only be a line or two of difference.

kthelgason:
Any update on this PR @smeevil ? If you need any help I'd be glad to jump in. I'm trying to land this so as to improve dogma support in Emacs' [flycheck](http://www.flycheck.org/en/latest/index.html)

smeevil:
@kthelgason I would be very happy if you could jump in :) after using spacemacs for about two months, I reverted back to the Intelli-J lineup of IDE's since I missed the deeper code integration / analyzation and other tools.

lpil:
I'm closing this due to inactivity. Reopen if you want to continue with this :)

PR #196, opened 2016-03-02
Read config from config/dogma.exs if not found then try ~/.dogma.exs

smeevil:
(This is a recreate from a previous pull request, but now properly scoped in its own branch)

It will now look if it can find config/dogma.exs in the project,
if not it will check if it can find ~/.dogma.exs. otherwise use the defaults

In this way a projects dogma config will always override the user config.

As this is file based, I'm not sure how to write a test for it
I did test this manually though using all four options being :

No config file at all
A config in config/dogma.exs
A config in ~/.dogma.exs
A config in config/dogma.exs and ~/.dogma.exs where the one in config wins over the home dir one
Hope you'll find this useful :)

### DIFF
--- a/lib/mix/tasks/dogma.ex
+++ b/lib/mix/tasks/dogma.ex
@@ -7,7 +7,7 @@ defmodule Mix.Tasks.Dogma do
   alias Dogma.Config
   alias Dogma.Reporters
 
-  #loads a config file with fallback chain. First found will be loaded.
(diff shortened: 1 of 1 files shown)


kthelgason:
@lpil What do you want to do to move this forward? I think it's a great improvement. It's pretty difficult to test this in any meaningful way. If you have some input I'd gladly contribute 😃

Issue #185, opened 2016-02-25
New rule: snake_case file names


lpil:

smeevil:
https://github.com/lpil/dogma/pull/198

PR #198, opened 2016-03-02
added a test to make sure all files are using snake_case

smeevil:
https://github.com/lpil/dogma/issues/185

### DIFF
--- a/lib/dogma/rule/snake_case_filename.ex
+++ b/lib/dogma/rule/snake_case_filename.ex
@@ -16,23 +16,14 @@ defrule Dogma.Rule.SnakeCaseFilename do
   a.valid.filename.ex
   a-valid-filename.ex
   """
+  alias Dogma.Util.Name
 
   def test(_, %{path: path}) do
     {dirname, filename, ext} = deconstruct_path(path)
-
-    desired_filename = filename
-      |> String.replace(~r/([A-Z])/, "_\\g{1}")
-      |> String.replace(~r/\W/, "_")
-      |> String.replace(~r/\A_/, "")
-      |> String.replace(~r/[_]+/, "_")
-      |> String.downcase
-
-    desired_path = reconstruct_path(dirname, desired_filename, ext)
-
-    if String.equivalent?(path, desired_path) do
+    if Name.snake_case?(filename) do
       []
     else
-      [error(path,  desired_path)]
+      [error(path, suggested_snake_case_path(dirname, filename, ext))]
     end
   end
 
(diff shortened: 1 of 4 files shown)


lib/dogma/rule/snake_case_filename.ex
```diff
@@ -0,0 +1,63 @@
...
+    desired_filename = filename
+      |> String.replace(~r/([A-Z])/, "_\\g{1}")
+      |> String.replace(~r/\W/, "_")
+      |> String.replace(~r/\A_/, "")
+      |> String.replace(~r/[_]+/, "_")
+      |> String.downcase
+
+    desired_path = reconstruct_path(dirname, desired_filename, ext)
+
+    if String.equivalent?(path, desired_path) do
```
lpil: Where does this function come from? Seems undefined

lib/dogma/rule/snake_case_filename.ex
```diff
@@ -0,0 +1,63 @@
...
+  aValidFilename.ex
+  a valid filename.ex
+  a.valid.filename.ex
+  a-valid-filename.ex
+  """
+
+  def test(_, %{path: path}) do
+    {dirname, filename, ext} = deconstruct_path(path)
+
+    desired_filename = filename
```
lpil: There is already a util module for attempting to detect whether a name is snake_case, use that instead.

If you can demonstrate that this implementation is better you can replace the implementation with this one :)

lpil:
Cool! A few comments above.

test/dogma/rule/snake_case_filename_test.exs
```diff
@@ -0,0 +1,23 @@
+defmodule Dogma.Rule.SnakeCaseFilenameTest do
+  use RuleCase, for: SnakeCaseFilename
+
+  should "not generate errors for snake cased files" do
+    result = Script.parse!("", "this_is_correct.ex")
+
```
lpil: Remove these newline please

test/dogma/rule/snake_case_filename_test.exs
```diff
@@ -0,0 +1,23 @@
...
+    assert expected_error == Rule.test(@rule, result)
+  end
+
+  should "error when a file contains invalid characters" do
+    result = Script.parse!("", "ThisIs-so.Wrong.ex")
+
+    expected_error = [%Dogma.Error{
+      line: 0,
+      message: "Rename ThisIs-so.Wrong.ex to this_is_so_wrong.ex",
```
lpil: Should we test each edge case separately?

lib/dogma/rule/snake_case_filename.ex
smeevil: not sure which function you mean ?

lib/dogma/rule/snake_case_filename.ex
smeevil: Using the internal tool for detecting snake case now.

test/dogma/rule/snake_case_filename_test.exs
smeevil: We could do that, but I don't really see the added value to that.
If you like give me a few cases you would like to test and I'll add them.

PR #207, opened 2016-05-01
Fix LiteralInCondition for tuples with variables

### LINKED ISSUE #204: LiteralInCondition: Literal value found in conditional in case statement
```
case { weather, weekday } do
  { :raining, :weekend } ->
    IO.puts "I'm will eat 2 apples!"

  { :raining, _ } ->
    IO.puts "I'm eating apple"

  { :sunny, _ } ->
    IO.puts "I'm happy!"

  { _, :sunday } ->
    IO.puts "I'm eating rice"

  { _, _ } ->
    IO.puts "I don't know what I'll eat"
end
```

Above code gives `LiteralInCondition: Literal value found in conditional` error. This is may be because `dogma` thinks `{ weather, weekday }` as Literal, but it has values inside. I think its an issue with `dogma`

Reference: http://stackoverflow.com/a/21079882

mgartner:
Fixes #204.

### DIFF
--- a/lib/dogma/util/ast_node.ex
+++ b/lib/dogma/util/ast_node.ex
@@ -20,10 +20,10 @@ defmodule Dogma.Util.AST do
   def literal?(x) when is_atom(x), do: true
   def literal?(x) when is_binary(x), do: true
   def literal?(x) when is_boolean(x), do: true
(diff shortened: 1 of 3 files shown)


lpil:
Great! Thank you.

PR #211, opened 2016-05-07
Clean up FunctionParentheses rule

mgartner:
I realized I could refactor my original code here by simplifying it into two functions that match when there is a rule being broken. I think this code is easier to understand.

Any feedback on how to format these functions with headers that exceed the 80 character line length limit? Let me know what you prefer.


lpil:
This looks great :)

The way you've formatted the function arguments seems clear and sensible to me,

PR #212, opened 2016-05-07
New rule: InfixOperatorPadding

mgartner:
Related to #165.

I think I finally got a solid implementation for this rule. It checks for padding around all infix operators, but handles `-` and `/` specially to allow things like `a = -1` and `&Thing.run/2`.

There's no way to get this rule working for versions of Elixir less than 1.1, because column information is not provided by the tokenizer. I still think this rule should be included in Dogma, so I've disabled it from running if the version is less than 1.1. If there is a cleaner way to disable this rule for certain versions that fits more in line with Dogma's rule configs, let me know.

### DIFF
--- a/lib/dogma/rule/infix_operator_padding.ex
+++ b/lib/dogma/rule/infix_operator_padding.ex
@@ -1,0 +1,156 @@
+use Dogma.RuleBuilder
+
+defrule Dogma.Rule.InfixOperatorPadding,
+  [fn_arrow_padding: false, elixir: ">= 1.1.0"] do
+  @moduledoc """
+  A rule that ensures that all infix operators, except the range operator `..`,
+  are surrounded by spaces.
+
+  This rule is only enabled for Elixir v1.1 or greater.
+
+  Good:
+
+      foo = bar
+
+      foo - bar
+
+      foo || bar
+
+      foo |> bar
+
+      foo..bar
+
+  Bad:
+
+      foo=bar
+
+      foo-bar
+
+      foo||bar
+
+      foo|>bar
+
+  By default, no space is required between the `fn` and `->` of an anonymous
+  function. A space can be required by setting the `fn_arrow_padding` option to
+  `true`.
+  """
+
+  @operators [
+    :comp_op,
+    :comp_op2,
+    :dual_op,
+    :mult_op,
+    :two_op,
+    :arrow_op,
+    :rel_op,
+    :rel_op2,
+    :and_op,
+    :or_op,
+    :match_op,
+    :in_match_op,
+    :assoc_op,
+    :stab_op,
+    :pipe_op
+  ]
+
+  @subtracters [
+    :number,
+    :identifier,
+    :")",
+    :")"
+  ]
+
+  @ignore_ops [
+    :-,
+    :..
+  ]
+
+  def test(rule, script) do
+    script.tokens
+    |> Enum.map(&normalize_token/1)
+    |> check_operators(rule)
+  end
+
+  defp normalize_token({a, {b, c, d}}), do: {a, b, c, d, nil}
+  defp normalize_token({a, {b, c, d}, e}), do: {a, b, c, d, e}
+  defp normalize_token({a, {b, c, d}, e, _}), do: {a, b, c, d, e}
+  defp normalize_token({a, {b, c, d}, e, _, _}), do: {a, b, c, d, e}
+
+  defp check_operators(tokens, rule, acc \\ [])
+
+  defp check_operators([], _rule, acc), do: Enum.reverse(acc)
+
+  defp check_operators([
+    {token, line, _, _, _},
+    {:identifier, line, _, _, _},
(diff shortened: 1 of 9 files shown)


lib/dogma/rule/infix_operator_padding.ex
```diff
@@ -0,0 +1,126 @@
...
+  end
+
+  defp check_operators([
+    {token1, line, _, end1, _},
+    {:dual_op, line, start2, end2, :-},
+    {token3, line, start3, _, _}
+    | rest], acc)
+  when (token1 in @subtracters or token1 == :")")
+  and (token3 in @subtracters or token3 == :"(")
+  and (end1 == start2 or end2 == start3) do
```
lpil: You can assign these value to the same variable you don't need to do the equality check.

i.e. `def foo(a, a)` is the same as `def foo(a, b) when a == b`

lib/dogma/rule/infix_operator_padding.ex
```diff
@@ -0,0 +1,126 @@
...
+    check_operators(rest, [error(line) | acc])
+  end
+
+  defp check_operators([
+    {token1, line, _, column, value},
+    {token2, line, column, _, _}
+    | rest], acc)
+  when token1 in @operators and not value in @ignore_ops and token2 != :eol do
```
lpil: Could we use metaprogramming to create a definition for each operator so we don't need to iterate over this @operators list every time.

test/dogma/config_test.exs
```diff
@@ -23,7 +23,8 @@ defmodule Dogma.ConfigTest do
     TemporaryEnv.delete( :dogma, :rule_set ) do
       TemporaryEnv.delete( :dogma, :override ) do
 
-        assert Dogma.RuleSet.All.rules == Config.build.rules
+        assert Enum.sort(Dogma.RuleSet.All.rules) ==
+          Enum.sort(Config.build.rules)
```
lpil: Why is this needed?

test/dogma/config_test.exs
```diff
@@ -67,7 +68,7 @@ defmodule Dogma.ConfigTest do
         override: %{ ModuleDoc => [enabled: false] }
       ) do
 
-        stdout = ExUnit.CaptureIO.capture_io(fn->
+        stdout = ExUnit.CaptureIO.capture_io(fn ->
```
lpil: I would like it if we could allow `fn->`

lib/dogma/rule/infix_operator_padding.ex
mgartner: 👍

lib/dogma/rule/infix_operator_padding.ex
mgartner: Sure thing, I'll work on that.

test/dogma/config_test.exs
mgartner: The order of the rules returned by `Config.build.rules` changed when a 33rd rule was added because `Config.get_rules` builds a list of rules by pushing them into a map, and then calling `Dict.values`. `Dict.values` doesn't guarantee the order of the values returned. My guess is that maps have different internal representations of their values when growing from 32 to 33 key/value pairs which changes the ordering.

test/dogma/config_test.exs
mgartner: Is this style used commonly in Elixir? I've only ever seen things like `fn -> ...` and `fn(x) -> ...`. Even in this repo,  These forms seem to appear more often than `fn->`.

test/dogma/config_test.exs
lpil: Interesting. Makes sense to me :)

test/dogma/config_test.exs
lpil: It's less common, but I like it.

mgartner: 👍  I can add an option in to allow this form, but I'm thinking enforcing it would be the responsibility of another rule, similar to the range operator `1..2`. I allow no spaces around this operator right now, but I don't enforce it. Do you agree?

lpil: Aye, that's the behaviour I would like.

rrrene:
@mgartner Nice work!

Obviously, I haven't been contributing to Dogma since my own little project started, but I want to give my 2 cents on this one, since I really like the fact that Dogma will have checks which require Elixir `> v1.0`.

What would you think about providing a way to define the Elixir requirement for a Rule in a consistent way. Returning `[]` in the `test` function seems a bit pragmatic. Maybe every check has to define its minimal version requirement, with the default pointing to `1.0`. This way you would know how many rules were tested, how many produced "real" errors and how many were skipped due to version conflicts.

Just a suggestion. Keep up the good work! :+1:

lpil:
I think that sounds like a grand idea. :)

mgartner:
@lpil I addressed all your comments, lmk what you think.

lpil:
Great! Thank you.

PR #213, opened 2016-05-28
Fix typo in function clause in json reporter

kthelgason:
This error was causing the json reporter to not work at all.
Unfortunately this wasn't caught by the tests so this commit
includes a small change to the tests that make sure this won't
happen again.


lpil:
Thank you.

Tests are failing!

kthelgason:
Sorry about that @lpil, fixed and force-pushed.

lpil:
Great! Thanks :)

PR #214, opened 2016-06-04
Add minimum_version option

mgartner:
Adds the ability to specify a minimum version of Elixir for a rule. The minimum version defaults to `>= 1.0.0`. 

Example:

```elixir
defrule Dogma.Rule.LineLength, [max_length: 80, elixir: ">= 1.1.0"] do
  # ...
end
```

@lpil @rrrene I threw this together, was this sort of what you had in mind based on your comments from #212?

### DIFF
--- a/lib/dogma/rule_builder.ex
+++ b/lib/dogma/rule_builder.ex
@@ -41,6 +41,8 @@ defmodule Dogma.RuleBuilder do
 
   defmacro defrule(name, opts, [do: module_ast]) when is_list(opts) do
     opts = [{:enabled, true} | opts]
+    opts = Keyword.put_new(opts, :elixir, ">= 1.0.0")
+
     quote do
       defmodule unquote(name) do
         alias Dogma.Error
--- a/lib/dogma/config.ex
+++ b/lib/dogma/config.ex
@@ -23,7 +23,10 @@ defmodule Dogma.Config do
   end
 
   defp get_rules do
-    rules     = Application.get_env( :dogma, :rule_set, @default_set ).rules
+    rules =
+      Application.get_env( :dogma, :rule_set, @default_set ).rules
+      |> Enum.filter(fn rule -> Version.match?(System.version, rule.elixir) end)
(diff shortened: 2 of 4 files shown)


rrrene:
:clap: Great!

One suggestion: I'd rename the option to `elixir_version` (or `elixir` if you want to be consistent with `mix.exs`) and use a **VersionRequirement rather than a minimum Version**:

    defrule Dogma.Rule.LineLength, [max_length: 80, elixir: ">= 1.1.0"] do
      # ...
    end

This would enable us to have Rules that e.g. only apply to Elixir 1.x.x but not Elixir 2.x.x or maybe you have a Rule that makes sense in Elixir 1.0-1.2 but no longer in Elixir 1.3. VersionRequirements as opposed to a minimal version would provide this flexibility but *not* introduce additional syntax, as they are already widely used in Elixir and implemented in the stdlib.

What do you think?

mgartner:
@rrrene makes sense to me. updated.

lib/dogma/rule_builder.ex
```diff
@@ -41,6 +41,8 @@ defmodule Dogma.RuleBuilder do
 
   defmacro defrule(name, opts, [do: module_ast]) when is_list(opts) do
     opts = [{:enabled, true} | opts]
+    opts = Keyword.put_new(opts, :elixir, ">= 1.0.0")
```
lpil: Would `:elixir_version` be more clear?

rrrene: `:elixir_version` and `:elixir` are both fine. the latter has the slight advantage that the corresponding key in `mix.exs` is called `:elixir`. Since Dogma uses Mix.Config I thought you might want to keep consistent naming here. :grin:

lpil: Seems reasonable.

lpil:
Looks good to me. Happy for me to merge?

mgartner:
👍

PR #216, opened 2016-06-09
Make Dogma available as an escript executable

kthelgason:
This adds support for creating an escript executable that can be make available globally, rather
than on a per-project basis. This is necessary in order for flycheck integration to work properly, among other things. See https://github.com/flycheck/flycheck/pull/969 for more info.


lpil:
Neat. Thank you :)

PR #217, opened 2016-06-10
Add mix app to `applications` in mix.exs

kthelgason:
This fixes a bug where Mix.Tasks.Loadconfig is not available
when running from the escript executable.


lpil:
Thank you

PR #222, opened 2016-08-28
Allow multiple '#' characters before a comment

paulfioravanti:
Currently, Dogma raises a `CommentFormat` error for comments that have multiple '`#`' characters before the beginning of the comment eg `## Currently invalid comment`.

This PR modifies the `get_comment` regex to look for one or more `#` characters in succession before the beginning of the comment.

Points of reference in favour of allowing multiple `#` characters before a comment:
- When generating a new Phoenix app, in the `web/channels/user_socket.ex` file, for example, you'll see comments starting with `##` generated from [its template](https://github.com/phoenixframework/phoenix/blob/master/installer/templates/new/web/channels/user_socket.ex)
- Rubocop's leading comment space cop [allows for multiple `#` characters](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/leading_comment_space.rb#L17)

### DIFF
--- a/lib/dogma/util/comment.ex
+++ b/lib/dogma/util/comment.ex
@@ -20,7 +20,7 @@ defmodule Dogma.Util.Comment do
   end
 
   def get_comment( {n, content}, acc ) do
-    ~r/#(?<content>.*)\z/
+    ~r/#+(?<content>.*)\z/
     |> Regex.named_captures( content )
     |> case do
       nil -> acc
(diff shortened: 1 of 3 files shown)


lpil:
Hi! Thanks for you contribution.

If we're adding to the behaviour it should be configurable. Add an option to switch between one space and at least one space. I'm happy for more than one to be the default behaviour.

axelson:
I just re-implemented this in #251 that adds an option to turn it on or off. Because of the requirement to have an option a recursive function is used instead of changing the `get_comment` regex.

PR #226, opened 2016-10-01
Allow pipe chains to start with a zero-arity function call.

### LINKED ISSUE #225: Function Pipe Chains must start with a bare value ... but how?
Dogma is giving me an error on the following line ...

```elixir
    conn = build_conn() |> get("/confirm?" <> @valid_link)
```

... but I have no idea how to convert that empty value to a bare value.

Is that even possible? If so, Dogma should suggest how to do it. If not, the error should be retracted.

scouten:
Fixes #225.


lpil:
This is great! Thank you for this :)

I'll release a new version with this fix immediately.

PR #232, opened 2016-11-01
Comments: Ignore spacing requirent for shebang line

### LINKED ISSUE #231: Comment rules should ignore shebang line in scripts
Unfortunately, including a space after the comment renders the standard shebang line inoperable:

``` bash
#!/usr/bin/env elixir 
```

I wonder if dogma should know to ignore the first comment line if it matches the `#!` pattern.

cgiffard:
* The unix shebang line requires the second character to be an exclamation mark, conflicting with the requirement for all comments to be followed by a space.
* This change adds an exception for comments without a leading space provided they occur on the first line of the script, and are immediately followed by an exclamation mark/bang.
* Tests added to support this.

Fixes #231.

### DIFF
--- a/lib/dogma/rule/comment_format.ex
+++ b/lib/dogma/rule/comment_format.ex
@@ -26,6 +26,12 @@ defrule Dogma.Rule.CommentFormat do
       "" ->
         errors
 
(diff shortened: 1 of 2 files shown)


lpil:
Very nice. Thanks!

PR #233, opened 2016-11-14
fix spelling and add little documentation for module

techgaun:
First, thanks for dogma. I use it a lot and has helped me to write better code in Elixir. Its just a spelling fix and little more addition of what dogma is in moduledoc


lpil:
Good change, thanks!

PR #234, opened 2016-11-14
only ignore dogma in root

techgaun:
`test/dogma` was being hidden/ignored by my IDE setting.. thought would be more appropriate to do this.


lpil:
Much better. Thanks :)

PR #237, opened 2017-02-10
Remove warnings shown by Elixir 1.4.1

JonRowe:
I noticed various warnings being showed on Elixir 1.4.1 when compiling dogma, as I strive to be warning free I thought I'd take a shot at removing them, most should be uncontroversial but let me know if you think theres a better way to tackle them?

I've broken the commits up to show whats been fixed where.

### DIFF
--- a/test/dogma/runner_test.exs
+++ b/test/dogma/runner_test.exs
@@ -24,6 +24,6 @@ defmodule Dogma.RunnerTest do
     """ |> Script.parse("")
     refute script.valid?
     results = Runner.run_tests( script, All.rules )
-    assert [%Error{ line: 1, message: _, rule: SyntaxError }] = results
+    assert [%Error{ line: _, message: _, rule: SyntaxError }] = results
   end
 end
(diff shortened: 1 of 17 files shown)


test/dogma/runner_test.exs
```diff
@@ -24,6 +24,6 @@ defmodule Dogma.RunnerTest do
     """ |> Script.parse("")
     refute script.valid?
     results = Runner.run_tests( script, All.rules )
-    assert [%Error{ line: 1, message: _, rule: SyntaxError }] = results
+    assert [%Error{ line: _, message: _, rule: SyntaxError }] = results
```
lpil: I would like to assert that the line number is correct. Is it not now?

lpil:
This is great, thank you. I've left one little note above.

test/dogma/runner_test.exs
JonRowe: It's line "0" for me

lpil: Huh, it should never be 0. Perhaps this is a bug.

lpil:
Thanks!

JonRowe:
Thank you!

PR #238, opened 2017-02-22
Add one more space to format example correctly

lee-dohm:
Example code was showing as plain text instead of as code because it had only three spaces before the code instead of four.

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -292,7 +292,7 @@ interpolation.
 
 Good:
 
-    output = inspect(self)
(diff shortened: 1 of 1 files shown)


lpil:
Tests are failing @lee-dohm https://travis-ci.org/lpil/dogma/jobs/204345771

:)

lee-dohm:
@lpil Updated ... thanks for the reminder :+1:

lpil:
Thanks :)

Issue #242, opened 2017-04-15
taken names rule triggers inside `defimpl`


P-Seebauer:
I wrote a data structure and added an implementation for the Inspect protocol. now dogma is rasing an error, because inspect is already defined.

Minimal example:
```elixir
defmodule Foo do
  @moduledoc "Some Foo data"

  defstruct data: nil

  defimpl Inspect do
    import Inspect.Algebra
    def inspect(%Foo{data: data}, opts) do
      concat ["#Foo<", to_doc(data, opts), ">"]
    end
  end
end
```
This rasies the error `inspect is already taken and overrides the standard library [TakenName]`.

I'm not sure if `inspect` is the only keyword for that a protocol exists, but the warning is very disturbing.

lpil:
Dupe of https://github.com/lpil/dogma/issues/201

What do you think the behaviour should be?

P-Seebauer:
Maybe it should not trigger inside `defimpl`? I'm not very accustomed with the language yet, but it seems like this is the way to use protocols, and if I want to cater one that's in Kernel, I'll get that message.

lpil:
Sounds sensible. Would you like to have a go? :)

P-Seebauer:
I'll take a look, but it might change the rule implementation fundamentally.

Also, if someone uses `@defdelegate` inside implementations (e.g. for `to_string`) it still seems odd.

```elixir
defmodule ACME.Name do
  defstruct first: "", second: ""

  def to_string %__MODULE{first: first, second: second} do
    Enum.join " ", [first, second]
  end

  defimpl String.Chars do
    defdelegate to_string(name), to: ACME.Name
  end
end
```

This would make `ACME.Name.to_string/1` accessible from outside (like it is with `Atom.to_string/1`): But inside the `ACME.Name` module, you can't use Kernel's `to_string` in bare form (and the compile fails if you try) anymore.

lpil:
It shouldn't be too hard to add a function clause that when a `defimpl` is encountered the AST is truncated to prevent any further traversal. Effectively preventing checking inside the `defimpl`. :)

P-Seebauer:
I wish I had seen this more early, now I'm deep in `Macro.traverse/4`…

Is there a reason, why the list is written by hand? It seems to me that it could be constructed automatically with something like 
```elixir
for what <- [:functions, :macros], descriptor <- Kernel.__info__(what) do 
  {name, _arity} = descriptor`
  name
end
```

[...]

PR #243, opened 2017-04-16
Allow implementation of base protocols in the takenName rule

P-Seebauer:
The takenName rule now doesn't raise an error for definitions inside protocol
implementations if the expected function name is a taken name.

see also #242

### DIFF
--- a/lib/dogma/rule/taken_name.ex
+++ b/lib/dogma/rule/taken_name.ex
@@ -44,11 +44,13 @@ defrule Dogma.Rule.TakenName do
     [:String, :Chars] => :to_string
   }
 
-  @spec reserved_words :: MapSet.t
-  def reserved_words do
-    @protocol_with_reserved_words
+  @reserved_words @protocol_with_reserved_words
     |> Map.values()
     |> Enum.into(@other_reserved_words)
+
+  @spec reserved_words :: MapSet.t
+  def reserved_words do
+    @reserved_words
   end
 
   @spec protocol_with_reserved_words :: %{required([atom]) => atom}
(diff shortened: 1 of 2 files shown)


lib/dogma/rule/taken_name.ex
```diff
@@ -21,59 +21,102 @@ defrule Dogma.Rule.TakenName do
...
   end
 
-  defp check_name(_, _, ast, errors) do
-    {ast, errors}
+  defp check_name(_, _, ast, acc) do
+    {ast, acc}
+  end
+
+  defp maybe_alter_words(reserved, impl_name, fun) do
+    if Map.has_key?(protocol_with_reserved_words(), impl_name) do
```
lpil: Rather than looking in the map twice why not use `Map.fetch` and pattern matching on the results with `fetch`? :)

lib/dogma/rule/taken_name.ex
```diff
@@ -21,59 +21,102 @@ defrule Dogma.Rule.TakenName do
...
+    [:String, :Chars] => :to_string
+  }
 
   @spec reserved_words :: MapSet.t
   def reserved_words do
-    @reserved_words
+    @protocol_with_reserved_words
+    |> Map.values()
+    |> Enum.into(@other_reserved_words)
+  end
```
lpil: How about we turn this into a module attribute? Saves us computing this value every time it is used.

lib/dogma/rule/taken_name.ex
lpil: Also I find the name of this function a little unclear. What is it doing?

lib/dogma/rule/taken_name.ex
P-Seebauer: Oh, I thought that is what that did… as I said not too accustomed, yet.

lpil:
Thank you!

Issue #248, opened 2017-10-06
Phoenix clean install 143 files, 835 errors


pmelnichuk:
disclaimer: I'm an elixir noob. New to the elixir community but trying to maintain a clean repo from the start.

Looks like phoenix from the base install fails miserably. is this normal?

Generated dogma app
Inspecting 143 files.

XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX...XXX.X...X.XXX...X....X..X...XX.XXX.X.X.XXXXX.X.XXXXXX.XXX.X.XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX

143 files, 835 errors!

lpil:
Phoenix doesn't adhere to the style specified by Dogma, however that does seem like a lot of files. Is it checking your deps, perhaps?

rhnonose:
It's suddenly checking inside `assets` folder (I recall not checking before, not sure).

lpil:
We have always checked any director called `assets` https://github.com/lpil/dogma/blob/426845e14cdb11a06561702c0a0fd61578205f6e/lib/dogma/script_sources.ex#L9

The difference now is likely that Phoenix 1.3 creates a directory called `assets` with NPM deps in it.

I would welcome a pull request that made Dogma ignore the `node_modules` directory.

PR #250, opened 2017-10-24
Ignore assets/node_modules .

rhnonose:
Addresses issue #248 

One problem that I found is that the tests are a little bit off.
`script_sources_test.exs` uses the `test/fixtures/app` to test, but when getting the path through `@fixture_path`, the line `|> Enum.reject( &String.starts_with?(&1, @ignored_dirs) )` in `script_sources.ex` doesn't match the paths to be ignored because it comes with the `test/fixtures/app` at the beginning, so it's not "properly tested".
I'm not sure how to refactor the tests to reflect that.

### DIFF
--- a/lib/dogma/script_sources.ex
+++ b/lib/dogma/script_sources.ex
@@ -9,6 +9,7 @@ defmodule Dogma.ScriptSources do
   @ignored_dirs ~w(
     deps/
     _build/
+    assets/node_modules/
   )
 
   @doc """
--- a/test/dogma/script_sources_test.exs
+++ b/test/dogma/script_sources_test.exs
@@ -24,7 +24,7 @@ defmodule Dogma.ScriptSourcesTest do
     end
 
     test "not return files that match given exclude patterns" do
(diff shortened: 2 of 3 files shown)


lpil:
Sorry I'm not quite understanding what you mean, could you run that by me again?

rhnonose:
@lpil the `@ignored_dirs` in `script_sources.ex` are not properly tested in `script_sources_test.exs`. If you remove either `_build` and now `assets/node_modules` from `@ignored_dirs`, the tests will still pass just fine.

This is because the `ScriptSources.find/1` rejects the paths with `String.starts_with?/2` and the tests use the fixture that is inside the test dir, making it having `test/fixtures/app` and the beginning, thus not matching the condition above.

rhnonose:
@lpil the `dep.ex` I added inside `assets/node_modules` is supposed to be ignored, but it's gonna break the build since the dogma self-check doesn't use the changes in the PR.
How can we fix this?

